PR #1, opened 2016-01-31
added kafka to rancher catalog

rawmind0:
Resubmiting pull request as you request me.

Added kafka + zk cluster to catalog. The clusters (zk and kafka) can scale and reconfigures itselfs based in rancher-metadata.


templates/kafka/config.yml
```diff
@@ -0,0 +1,8 @@
+name: Apache Kafka
+description: |
+  (Experimental) Kafka cluster
+version: 0.8.2-rancher1
+category: clustering
+release by: Raul Sanchez
+requires: rancher-0.47.0
```
cloudnautique: this field should be `minimum_rancher_version`

templates/kafka/config.yml
```diff
@@ -0,0 +1,8 @@
+name: Apache Kafka
+description: |
+  (Experimental) Kafka cluster
+version: 0.8.2-rancher1
+category: clustering
+release by: Raul Sanchez
```
cloudnautique: I don't think you can have spaces in the keys, can you add an '_'? Also, this isn't a valid field for our API, though I don't think it hurts to have it there. I know we were toying with maintainer, but I don't see that in the code.

cloudnautique:
Thanks for opening this PR! Couple comments in line.

templates/kafka/config.yml
rawmind0: i saw this field in other packages in your offical repo. Do you have docs about the rules that you say??

templates/kafka/config.yml
rawmind0: i don't understand your comments. i made the changes that you request and checks start to fail.  Add '_' where??
Do you have docs with these rules?? it would be easy if it exists.

rawmind0: Done the changes, but drone build fails. Is this issue about the changes or about the system??
The package is working well in my rancher.

cloudnautique:
Could you rebase master, and squash? I messed up the CI configuration, and it should now be more correct. Thanks!

cloudnautique:
Please squash. I ran the CI tests locally, and everything looked good there.

rawmind0:
squash done.

cloudnautique:
Sorry, we typically squash all our commits into a single commit. 

I ran the CI scripts manually, and it all worked. It looks like the validation directory didn't get into that merge.

I think if you roll back the last merge commit on your branch, and set this repo as a git remote, you should be able to:
git fetch rancherupstream
git rebase -i rancherupstream/master
git push -f origin master

Of course I recommend saving a copy of your directory, just in case it goes completely awry :) 

while on your branch. The `-i` will ask you which commits to keep and which to squash.

rawmind0:
Aleluya!!!! :P

templates/kafka/config.yml
```diff
@@ -0,0 +1,8 @@
+name: Apache Kafka
+description: |
+  (Experimental) Kafka cluster
+version: 0.8.2-rancher1
+category: clustering
+maintainer: "Raul Sanchez <rawmind@gmail.com>"
+minimum_rancher_version: rancher-0.56.0
```
vincent99: This needs to be `v0.56.0` to work

templates/kafka/0/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+zk:
```
vincent99: Is there a reason this needs a copy of its own zookeeper, apparently based on ours, instead of having a question of type service to allow the user to point to their own (possibly already existing) stack?

rawmind0: Zookeeper it's not a copy, neither it's not based in yours, i don't understand why you say it??. I build my own and i deploy it as you can see in docker-compose file. 
Reasons to kafka have it's own zookeeper:
First reason is that mine is about 130 Mb smaller than yours. :P
This is yours - rancher/zookeeper      3.4.6-2    312 MB
This is mine - rawmind/rancher-zk    3.4.6-1   182.4 MB

Second reason is that mine could scale and get reconfigured dinamically, with confd obtaining data from rancher-metadata v2. 

Third reason is that in mine you could configure memory sizing when you launch it.

Fouth reason is that kafka recommends have a dedicated zookeeper to work well in terms of performance.

templates/kafka/config.yml
rawmind0: Done, parameter modified.

templates/kafka/0/docker-compose.yml
vincent99: Sorry, I just read the image name and didn't go digging into it because I've had this discussion with apparently someone else before.  Still, those all seem like good things to put in a PR to the rancher-catalog zookeeper; we take PRs there too.  Or a rawminds-zookeper-thats-awesomer-than-ranchers template in here.

I  just don't like the idea of having a separate re-implementation of the same service in every template that happens to need common basic services like this.  If you make a, say, Solr template next is it going to have another copy of exactly the same zookeeper service in it?

cloudnautique: Sorry, the confusion around the copy was my fault.

rawmind0: Don't worry about that guys, it's only a misunderstand. 
Please review the package carefully and if you like it and you want to include zookeeper features in your package, i'm happy to make a new PR to rancher-catalog zookeeper, it's no problem..Or may be i think about that funny name to be in here... ;P

@vincent99 i understand your point, and almost i share your opinion, but i think that in this particular case, to have a particular instance of zookeeper is more than justified due to the kafka nature. In terms of scale and/or performance, to have a shared zookeeper may be a problem for kafka, and eventually for others apps that try to use the same shared zookeeper.

vincent99: The changes you described vs our ZK sound like they would be fine to me as a PR, but I'll let @cloudnautique review the specifics.  I have no issue with "Kafka should have its own dedicated instance of ZK".  The part I don't like is really not related to you specifically at all, but about services including separate dependent services in general.. sharing the definition of a service, not the deployed instances of it:

- Including it changes from "should have a dedicated" to "you **will** have a dedicated".. maybe I have a good reason to want to share with something else.

- I'm now tightly coupled to your version of ZK. I can't use the Rancher one, or someone-elses-even-awsomer-zookeeper template, or an External Service that points to the ZK-as-a-service my company runs, or a 3rd party one hosted elsewhere, or my own that has an extra option for a knob I need to tweak that you didn't think of.

- As the Kafka maintainer you now have to keep up to date with ZK releases and update the Kafka template every time there's a new one instead of letting someone who maybe cares about and keeps up with ZK more often do it.  As a user I have to worry about whether your version of ZK packaged in is up to date.

The alternative is you have a question with `type: service`, `description: "Choose the Zookeeper service to use.  It should really be dedicated just to Kafka, see $url for more info."`.  The main disadvantage to me is really that it's less convenient.. the user has to have already created one.  And also that it then shows up as a separate service even though it's (presumably) dedicated to Kafka.

I think we really need a couple more features to do something like nest catalog services and have a question for "choose an existing or deploy a new $service" that would allow you to reference ZK without including a new copy of it..

Anyway if you feel strongly it has to be a dedicated ZK then this is the only way to enforce that and we can merge this as-is.  But I would really prefer separating it out improving the Rancher zookeeper and/or including your own rawminds-awesome-zookeper here.

cloudnautique: @rawmind0, for now adding it to the community catalog would be good. The next release has a feature that will allow us to keep quorum in a ZK cluster during an upgrade, and we will be revisiting the template again. See rancher/rancher#2494. I'll be looking at your template features as we roll out the next iteration of the Rancher version.

In the mean time the community could benefit from the ZK image.

rawmind0: @vincent99. I'm completely agree with you. 

As i said before, i understand very well your point. I would shared it completely, if you already have a nested catalog that it could deploy an already defined service. When i started to think about how to build the package, i thought to do it exactly as you said, but i decided to change my mind, although i know it is a worse design , due to the missing feature that you say of the nested catalog. 

Are you working in that way?? Is in you backlog this feature?? You make so happy if you say that yes. :)

If this feature is coming on, i have no problem to do that you suggest. I do a new PR to your rancher-catalog zookeeper and i do an easy modification in confd templates for broker and push it again. 

Right now, zk cluster and broker clusters, can run in a dynamic or in a static mode. By default, they run in dynamic mode (you could scale up and down the cluster dinamically and separately), checking every $CONFD_INTERVAL seconds (60 by default) for new config, but you if you set $CONFD_INTERVAL to 0 they run in static mode (you have to reboot the clusters to scale). Do you think that the possibility to choose between the 2 run modes, is a valuable feature??

vincent99: Nobody is immediately working on nesting but yes I expect we will add a way to address this fairly soon.

Is there some advantage to the static mode?  It doesn't sound like it from the way you describe...

rawmind0: Guys, i made the changes that we talk about. 
Kafka now have a question type:service to select zookeeper stack/service.
Almost i add zookeeper service.

cloudnautique:
@rawmind0 , Why two versions of the Kafka template, 0 and 1? If a user were to deploy template 0 they would immediately be prompted to upgrade to version 1 of the template folders.

rawmind0:
Done

cloudnautique:
LGTM, @vincent99 ?

vincent99:
Not sure why README.md changed and would prefer squashing, but LGTM

cloudnautique:
@rawmind0, can you squash, and we will merge this in?

rawmind0:
Done

cloudnautique:
Awesome! Thank you.

PR #5, opened 2016-02-22
Update MongoDB image and let users enter their desired Version

dominikhahn:


templates/MongoDB/1/rancher-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+  minimum_rancher_version: v0.46.0
+  questions:
+    - variable: replset_name
+      description: "Name of the MongoDB replicaSet"
+      label: "ReplicaSet Name"
+      type: "string"
+      required: true
+      default: "rs0"
+    - variable: mongo_version
+      description: "Version of the MongoDB Image</a>"
```
vincent99: This has a random `</a>` in it, and should probably link to the [list of versions](https://hub.docker.com/_/mongo/).

Does mongo handle replicas with different versions of the software well?  Using a generic default like "3" is a recipe for ending up with varying versions.

dominikhahn: True, deleted that one.

About the mongo versions. MongoDB is able to use older DB Members but as a minimum of 3.0. The main reason of this pull request is to upgrade the mongodb image to 3.2.1 :-)

So could also be:
`image: mongo:3`
see [https://hub.docker.com/_/mongo/](Mongo Image)

PR #13, opened 2016-03-04
New templates: Janitor, Minecraft

sshipway:
Two new templates, tested on our Rancher 0.56 server.

**Janitor**
This uses the meltwater/docker-cleanup image to periodically remove old container images.  This is useful because many lightweight Rancher Hosts do not have much disk space available locally.  We run this on all of our Hosts.

**Minecraft**
This starts a Minecraft server with a number of configurable options, using the itzg/minecraft-server container image.  This is useful for demonstration purposes (Minecraft shows a non-http use of the load balancer, and the client can constantly display which services are reachable) and also for wasting lots of time playing.


Negashev:
@sshipway janitor kill "start once" containers, and data is lost
screenshot 62

sshipway:
This is correct; the janitor has no way to know if a container was a run-once that needs to be kept, or one that is no longer required.  What you can do is to exclude the image from the cleanup when you start up Janitor by using the exclude option.

PR #18, opened 2016-03-07
New template: DataDog

janeczku:
dd-hero-img

Based on the official [Datadog Agent image](https://www.github.com/DataDog/docker-dd-agent) with a few changes to the init script to better support the Rancher environment.

* Hostname reported to DataDog is the actual name of the host (per Rancher Metadata service)
* User can specify existing `host labels` which will be used to set DataDog key:value tags for the host
* User can either deploy the DataDog Agent globally on every host or deploy a single instance of a standalone DogStatsD
* The later is useful if one merely wants a StatsD aggregator to which other services can send metrics

To try it out add the following custom catalog URL on the Admin->Settings page:
> https://github.com/janeczku/rancher-catalog.git


cloudnautique:
Cool. Will merge as is.

As a suggestion for the future, you can use the upstream Datadog image, and just create a config sidekick with an exported volume. In the volume put your script, and in the compose entry for the Datadog container just add the entrypoint: /path/to/my/entrypoint.

Then a version change is likely an increment to the version number in the compose file vs. a container rebuild/push etc.

smw355:
Jan, would you be interested in writing a blog on the Rancher site about the new catalog item?

janeczku:
@smw355 Why, thank you for the suggestion! :blush: I am inclined to, but can't presently say if i find the time soon. Let me get back to you on that!

smw355:
Sounds good Jan, take your time!

Best Regards,

Shannon Williams
shannon@rancher.com<mailto:shannon@rancher.com>
+1 650-521-6902
www.rancher.com<http://www.rancher.com>

PR #36, opened 2016-03-16
Added gocd server and agent to the catalog

rawmind0:
Catalog packages to run, gocd server and scalable gocd agents.


cloudnautique:
Small nit, are there gocd svgs? We prefer those over png if available.

rawmind0:
Yes, i know your preferences in svg vs png, but i didn't find that logo in svg format.

cloudnautique:
We don't have an official check for image type at the moment, so I have to ask :)

 Thanks!

PR #45, opened 2016-03-21
add Cloudflare

LLParse:


cloudnautique:
keeping in rancher-catalog

PR #53, opened 2016-03-21
add kubernetes

LLParse:


cloudnautique:
Keeping out of catalog for now. The one needed for Rancher envs is bundled with rancher.

PR #56, opened 2016-03-21
Add ECR Credential Updater template.

johnrengelman:


templates/ecr/0/rancher-compose.yml
```diff
@@ -0,0 +1,24 @@
+.catalog:
+  name: "ECR Credential Updater"
+  version: "v1.0"
```
cloudnautique: This should line up with the version in config.yml (v1.0.0)

johnrengelman: Updated.

PR #69, opened 2016-03-23
etcd HA template

LLParse:
We will need to build rancher/etcd:v2.3.0 before this can be merged. The PR for that is here:
https://github.com/rancher/catalog-dockerfiles/pull/51

EDIT: image is built, ready for test/merge


LLParse:
See https://github.com/rancher/catalog-dockerfiles/pull/51 for v2.3.0 image

PR #72, opened 2016-03-24
Select LDAP attribute for  search  username

Rhoa:


cloudnautique:
@sshipway thoughts?

@Rhoa, we will need to squash the commits down before merging.

vincent99:
@cloudnautique There's a button in GitHub to squash and merge now, so the contributor doesn't really need to do it anymore.

sshipway:
@Rhoa - This would need to be done for the newly merged 1/ template (req #99 ) as well as the original 0/ template.  Sorry for moving the goalposts.

If you are going to parameterise the username attribute, is using an enum the right way to go?  Maybe just make it a text field - so as to support other attributes than cn and uid... people may have custom schemas or use the mail attribute as the username

cloudnautique:
@rhoa @sshipway what is the story with this PR? Should it be closed?

sshipway:
I think the PR should be extended to include the same changes in the newer 1/ template before merging?  No point in making changes only to the older template.

cloudnautique:
closing. Please open a new PR if you'd like to address issues.

PR #78, opened 2016-03-29
Port Spark template from Helm to K8s templates

cloudnautique:
This is a simple port of the helm template of spark.


LLParse:
LGTM!

PR #79, opened 2016-03-29
Add template validation, fix categories

LLParse:


integration/core/test_catalog.py
```diff
@@ -3,6 +3,8 @@
 import subprocess
 import sys
 import os
+import re
+# import yaml
```
cloudnautique: Is this needed?

LLParse: It will be when lines 117-119 get uncommented and we start validating the template yaml. I didn't do so because there are templates without any questions (but couldn't justify ripping it all out)

PR #86, opened 2016-03-30
TripleA dedicated servers

LLParse:
@cloudnautique @ibuildthecloud @vincent99 I built this template to discuss port conflicts. Here's the requirement I'm trying to address:

A container's public (host) and private (container) port must match. This is because a single port parameter is used to specify where the server binds _as well as_ where clients connect (advertisement). While this may not be the best design, it exists in lots of software. This presents a problem when trying to define a template which schedules multiple containers on a single host. There will be a port conflict.

The template below parameterizes the port number (although we don't care what it is) and adds a port # constraint. This allows us to deploy multiple stacks with a manually selected port number to achieve multiple servers on a single host.

I think the solution is to implement dynamic port reservations, or maybe there is a more elegant way to work around this?


rawmind0:
Hi @LLParse ,

I'm sorry, thank so much...

rawmind0:
Hi @LLParse , feel free to reopen if it's needed...

PR #87, opened 2016-03-31
Adding Rancher Server with Nginx proxy and Let's Encrypt SSL in catalog

KrustyHack:
Ay,

I hope my work is fine.

Regards,


djskinner:
Out of curiosity how does running a rancher server instance as a stack from a rancher catalog entry actually work? What about the existing instance of rancher server?

KrustyHack:
Ay,

I don't understand your question (french here :p). There is no problem if you deploy a rancher server from catalog to another server.

djskinner:
You understood fine since you answered my question perfectly. Thanks. :)

djskinner:
My only other suggestion would be not to use `latest` tags. It happened to me once before where the Prometheus catalog entry broke when a new version of one of the images was released. If the versions are locked down then this cannot happen.

sshipway:
I agree with the "no :latest tags" rule for templates.  You can always release a new template version if you want to support a later image version.

The "running Rancher in a Rancher stack" is the same chicken-and-egg issue we have with running the VMWare Virtual Centre in the VMWare cluster that it manages.  Yes, it can work, but you have to have a way to get it back up if everything is shut down -- and you can end up with  some gnary dependencies.

PR #94, opened 2016-04-03
Update docker-compose.yml

kalw:
updated to work w/ https://github.com/sshipway/Portus/pull/1


kalw:
Only valid when pull request will be merged =)

sshipway:
I think that this should be done as a new Template version rather than modifying the old, so that people can upgrade.  Also, it should probably give an explicit version for the sshipway/portus container tag for stability.  I can make a pull req for this once I get the new sshipway/portus container image built, tested and uploaded.

kalw:
sure. I'll look forward into it =)

PR #95, opened 2016-04-04
PXC catalog item

Flowman:
Percona XtraDB Cluster template for rancher with clustercheck


cloudnautique:
Can you squash to a single commit please.

PR #96, opened 2016-04-04
Add tcp socket for logstash 1.x

dominikhahn:


templates/logstash/1/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
+logstash-indexer-config:
+  restart: always
+  image: rancher/logstash-config:v0.2.0
+  labels:
+    io.rancher.container.hostname_override: container_name
+redis:
+  restart: always
+  tty: true
+  image: redis:3
```
LLParse: Why did we go from v3.0.3 in the previous template to v3 here?

LLParse:
One small nitpick, otherwise LGTM

templates/logstash/1/docker-compose.yml
dominikhahn: Static version added :+1:

PR #99, opened 2016-04-06
Updates for Janitor and Registry templates

sshipway:
Janitor
- switch back to using the official meltwater image now that the regexp changes have been accepted into that branch.  No longer any need to use my own forked and patched copy.

Registry 
- upgrade to registry 2.3.1
- New version of Portus container with recent changes pulled in, plus fixes for reported certificate issues
- Make certs filesystem mount rw for compatibility with registry 2.3.1


sshipway:
Thanks to @kalw for merging the Portus 2.0.4 updates in.

PR #100, opened 2016-04-07
Janitor needs neither privileged mode nor network access.

steigr:
- remove "privileged: true" property
- add "net: none" property

The cleanup tasks runs entirely without networking on /var/run/docker.sock. The calling process can access the docker so it does not need to have privileged mode.

Signed-off-by: Mathias Kaufmann <me@stei.gr>


sshipway:
As well as removing stopped containers and unused images, the Janitor process will also remove orphaned volumes.  This may be the reason for the Privileged requirement; in addition, some Docker installations may have permissions on /var/run/docker.dock that require privileged mode.

The documentation for the underlying docker-cleanup container recommends use of privileged mode, which is why it is set (though your system may not require it)

I agree with the use of "net:none" as it does not need any networking.

cloudnautique:
On the privileged issue, can we set it as a question of enum type. Default should be true to keep consistent with established behavior.

kkvilekval:
Trying to run janitor in new environment and am seeing this in the log:

```
time="2016-05-18T18:27:57Z" level=error msg="Stack Create Event Failed: Service 'cleanup' configuration key 'privileged' contains an invalid type, it should be a boolean." eventId=e6d80cde-cb06-4aa0-9acb-50ed937980f5 resourceId=1e4 
time="2016-05-18T18:28:42Z" level=info msg="Stack Create Event Received" eventId=c7b3b1ac-0e0d-4ec2-bbc8-77574c77361b resourceId=1e4 
time="2016-05-18T18:28:42Z" level=error msg="Could not parse config for project janitor1 : Service 'cleanup' configuration key 'privileged' contains an invalid type, it should be a boolean." 
time="2016-05-18T18:28:42Z" level=error msg="Stack Create Event Failed: Service 'cleanup' configuration key 'privileged' contains an invalid type, it should be a boolean." eventId=c7b3b1ac-0e0d-4ec2-bbc8-77574c77361b resourceId=1e4 
```

It seems to have the issue whether I set the privileged or not.

sshipway:
This is a problem caused by a bug introduced in the latest Rancher versions, where booleans parameters in the rancher-compose cannot be specified using template variables as the syntax check is run before the substitution: https://github.com/rancher/rancher/issues/4822
Currently the only way around this is to set up your own version of the template or set up the stack manually.  See the rancher issue (link above) for progress.

PR #102, opened 2016-04-09
Add scalable Hazelcast template

LLParse:
Features:
* Multi-node deployment
* Scale up/down

This is a WIP and not ready to merge.


cloudnautique:
@LLParse what would you like to do here?

PR #111, opened 2016-04-18
add Sematext Docker Agent

megastef:


pianomansam:
I just saw this and tried it out today. Unfortunately, it looks like the part that configures the io.rancher.scheduler.global label doesn't properly set it to true. I created a stack and it only populated to one host. Looking at the configuration for the stack, it looks like this:

`io.rancher.scheduler.global: ''`

I believe it should look like this:

`io.rancher.scheduler.global: true`

megastef:
Thx for reporting this issue, should be fixed in https://github.com/rancher/community-catalog/pull/147

megastef:
@cloudnautique could you pls. merge https://github.com/rancher/community-catalog/pull/147 to fix the issue mentioned by @pianomansam. Thx.

PR #113, opened 2016-04-19
fix: gocd-agent docker-compose template syntax

rawmind0:
There is a syntax error in gocd-agent docker-compose.yml, defining external_links, you get Activating (Bad response statusCode [422]. Status [422 status code 422]. Body: [code=InvalidCharacters, fieldName=name] from [http://localhost:8080/v1/projects/1a5/services/1s45/?action=setservicelinks])

Solved with this PR.


rawmind0:
added trafeik catalog package

PR #114, opened 2016-04-20
update ECR Credentials Updater image to v1.0.1

smazurov:
### What is the problem

When trying to set up `ECR` on Amazon Web Services through rancher, I was running into rancher/rancher#3634 and rancher/rancher#4250. I went to explore if the ECR Credentials Updater got a new version, and sure enough, it did. Doing a manual update, I was able to successfully have it update auth credentials and my ECR-stored repositories/images worked without any issues.

### How did it get fixed

Added v1 of ECR Credentials Updater which bumped the image from `1.0.0` to `1.0.1`

### How can someone test / see it ?

Manually grab [`docker-compose.yml`](https://github.com/smazurov/community-catalog/blob/66eafe04f63a46ac23f7159bf4310e383c5f7e11/templates/ecr/1/docker-compose.yml) and add it as a service in Rancher. Make sure to replace variables that are typically coming from the `.catalog` definition.


smazurov:
closing since there is another pr open that accomplished the same thing #117

PR #121, opened 2016-04-26
Update CloudFlare to use rancher/external-dns:v0.4.0

sdurrheimer:
Requested in rancher/external-dns#32

@alena1108


alena1108:
LGTM

PR #125, opened 2016-05-02
Added RabbitMQ stack with clustering

ddmng:
I added a RabbitMQ 3.6.1 stack that can scale using Rancher UI "Scale" commands.
It's based on a rabbitmq:3.6.1-management image and includes a load balancer that manages the 15672 and 5672 ports.
The base image starts a supervisord daemon with two services, one for the rabbit server and one that joins the container to the cluster if it's not the first one in the stack. To decide if it's the first or not, it looks at the rancher-metadata service comparing its name with the name of `/self/service/container/0/name`.
I'm making a pull request on catalog-dockerfile as well.


andyshinn:
Shouldn't the join script be handled with a sidekick container so the official image doesn't need to be modified?

ddmng:
I'd like to keep the official image too but AFAIK to stop/join/start a RabbitMQ instance you should be on the same machine/container where the rabbit server is running. How can I start that commands sequence from another container like a sidekick?
I can check if some plugin can help.

andyshinn:
Have you checked out the Elasticsearch configuration sidekick? It seems that by default, sidekicks start before the main service (specifically so they can write out configuration that the main container can use). Then, you could just do a volumes from the configuration container.

Let me know if any help needed for testing. I'm having some requests for a RabbitMQ stack in our Rancher environments so I'm happy to give this or another method a go.

ddmng:
Yes, I used that way in most of my own stacks. In my latest tests with Rabbit I was trying to issue the `stop_app`, `cluster_join`, `start_app` from the sidekick's entrypoint command but the first rabbit instance was always rejecting connections. I linked this way because it wouldn't need any `-conf` image.

I'll try to do something closer to elastic-conf in the next days.

andyshinn:
I wonder if those commands just add relevant configuration to the file that could be grabbed from the metadata API and written out with confd instead?

ddmng:
You're right. In the config file manual I've found that it could work just
like elastic:
```
cluster_nodes
Set this to cause clustering to happen automatically when a node starts for
the very first time. The first element of the tuple is the nodes that the
node will try to cluster to. The second element is either disc or ram and
determines the node type.
Default: {[], disc}
```
Starting from the elastic rancher stack, I should do it in an easier way
than using the `rabbitmqctl` command.

Il lun 16 mag 2016, 19:26 Andy Shinn <notifications@github.com> ha scritto:

ddmng:
OK, I did it in the declarative way, like elastic. It seems to work both scaling up and down. The `rancher/rabbitmq-conf` in `catalog-dockerfiles` repo now runs confd instead of supervisord and it creates `/etc/rabbitmq/rabbitmq.config` file in the shared volume.
Could you test it a bit? I can't test on my devel environment so I used a couple of vagrant on my local notebook so I had issues related memory and CPU resources.
I removed the load balancer on 15672 port. Do you think it should be included? (...this should be a decision for all the stacks in the catalog).
I go and update the READMEs. Let me know of any issues.

ddmng:
I'll make user and pass configurable in a future PR...

ddmng:
@andyshinn I moved all in PR #151

PR #128, opened 2016-05-05
Add XPilot X-windows game template

sshipway:
This adds a new Community Catalogue item to run the XPilot game, with a server and multiple Clients running exclusively over the Rancher private network.

These applications are X-Windows clients, talking back to a workstation with X for their display.  This demonstrates how it is possible to use Rancher to host instances of an X application instead of having multiple remote hosts and accessing their displays via VNC or similar.

To make it work multiplayer, you just need to Clone the "client" service in the stack, and give it a different DISPLAY and NAME in the environment.


sshipway:
May be a problem, will resubmit

sshipway:
Seems to be a cross-host networking issue if Server is running CoreOS beta.  Compensated by adding soft host affinity for client

sshipway:
Have also added an update for the Registry template.  This updates to using the 2.0.5 container version of Portus (with a bugfix for multihomes containers) and adds some soft affinity rules to avoid some anecdotal networking issues reported under non-recreatable situations

PR #130, opened 2016-05-10
Add a default go server port

aisensiy:


templates/gocd-server/0/docker-compose.yml
```diff
@@ -1,4 +1,6 @@
 gocd-server:
+  ports:
+  - 8153:8153/tcp
```
cloudnautique: Can you make this a question in the rancher-compose.yml. Hard coded ports cause issues.

aisensiy: Ok, I will do that. Thanks for your feedback.

PR #144, opened 2016-05-17
Use question and answer to set public port for gocd server

aisensiy:


aisensiy:
@cloudnautique I opened a new pull request here :-).

PR #152, opened 2016-05-23
Zookeeper v3.4.8 upgradable, configured with confd

LLParse:


ruimarinho:
@LLParse how does zookeeper handle leader election during an upgrade?

cloudnautique:
@LLParse is this still needed?

PR #154, opened 2016-05-24
upgrade janitor to latest image version, docker 1.9+ support

mfournier:


sshipway:
Nice for new support.  Can we make the 'keep containers named' default to known run-once containers in the Rancher catalogue (parts of convoy and so on)?

I think it might be necessary to remove the 'privileged mode' option from this new template version, as it has been shown to be incompatible with a lot of docker/rancher versions.  Default it to 'true' to maximise compatibility.  A separate pull req is in place to remove it from template v1.6

cloudnautique:
@mfournier @sshipway whats the status of this?

sshipway:
If I remember correctly, there was a problem with parameterising PRIVILEGED_MODE as it breaks the syntax check (since the syntax check is done before the templating and '${PRIVILEGED_MODE}' is not 'true' or 'false' until expanded).  So, probably continue to hardcode this to 'true' rather than being parameterised.  Other than that looks OK.

I think the KEEP_CONTAINERS_NAMED parameter is a good idea and the new documentation, as well as the image version update.

cloudnautique:
@mfournier would you mind updating as @sshipway suggests?

mfournier:
@cloudnautique I just updated the patchset as suggested by @sshipway. Thanks for reviewing !

@sshipway, I'm a bit worried about your request for listing all the "known run-once containers". Wouldn't it be better to switch the default to `**All**` if we are to expect such containers by default ?

I also updated the docker-cleanup to the latest released version: https://github.com/meltwater/docker-cleanup/compare/1.7.2...1.8.0

sshipway:
@mfournier -- the current janitor template defaults the KEEPC parameter ("keep containers") to "*:*" which is the same as "**All**" (the '**All**' symbol is a recent addition that was not available at the time the template was first made).   So currently it has the same effect.  Defaulting to keep all containers is a good idea since, as you say, we don't know what run-once containers there might be -- and Rancher takes care of removing expired containers where necessary anyway.

sshipway:
Just a note, as I am now not so involved in Docker work, I'm happy for someone else to take over the lead on the Janitor template if they want.

mcanevet:
@cloudnautique would it be possible to merge this? The `meltwater/docker-cleanup` image before `1.7.2` does really annoying things by removing the folder of the volume without removing the volume from Docker. This prevent Docker from creating the volume again (because it thinks that it already exists), and it fails `no such file or directory` when trying to re-create a container that mounts the same named volume.

guewen:
I confirm the issue @mcanevet explained, it breaks my automated workflow for tests instances on Rancher as I `rm` a container which cannot be re-created due to that bug.

@cloudnautique @sshipway could you consider to merge?
Thanks!

sshipway:
I'm happy for this to go ahead.  As I mentioned before, I'm on other projects now, so cannot test as I no longer have access to the Rancher environment.

PR #162, opened 2016-05-30
Remove option to set Privileged mode, as this cannot be parameterised

sshipway:
The current template does not work with docker-compose.  This removes the configurability of the Privileged mode so that it will.  Note that hardened Docker setups will not be able to run Rancher without Privileged mode, and clearing of orphaned volumes requires it.


cloudnautique:
I'll merge, but I'm pretty sure default RH variants will not work without privileged. At least that has been our experience.

sshipway:
@cloudnautique - should we then set the privileged:true in the template (as originally done)?  There are too many variants of OS for me to fully test :(.  I'm quite happy for it to have privileged:true hardwired in, if you want to make the change.

PR #168, opened 2016-06-02
Adding Rancher Server with Nginx proxy and Let's Encrypt SSL in catalog

KrustyHack:
Ay,

I hope my work is fine.

Regards,


rawmind0:
Hi @KrustyHack ...

Thanks for the PR and to contribute and sorry about the delay.

I'm sorry, but i don't understand very well what is the target of this package.  You already need a rancher server to be able to launch any catalog package.

Otherwise, you could also use the built-in rancher lb or traefik package to do the same task.  

I close the PR. Please reopen it if i've missed something. 

Best regards...

KrustyHack:
Hi,

No problem. The target was to be able to deploy Rancher server bundle for my customers on there servers. And I thought it would maybe be usefull for other people. :)

Regards,

PR #170, opened 2016-06-06
Added Concrete5 CMS Template. with auto configuration in v0.2 or manua…

borisno2:
Concrete5 CMS template with Mariadb database backend. Setup with auto configuration via a sidekick container in v0.2 or manual config (ie via the standard CMS webpage) in v0.1.


cloudnautique:
would it be possible to get a single image for the icon.. it looks like there are three

borisno2:
Hi @cloudnautique 
Latest commit has single logo as a png.
Let me know if there are any other issues.
Thanks!

PR #173, opened 2016-06-07
optimize image file size using imageOptim

mastertinner:
i've reduced the file size of all bitmap images using [ImageOptim](https://imageoptim.com/mac)


cloudnautique:
Thanks @mastertinner.

I think doing something like this makes sense, but it would be ideal if it was part of the delivery process so that it is kept up to date. Given the current structure of the catalog, Im not sure the best way to go about it at the moment need to give it some more thought. 

Any ideas?

rawmind0:
Hi @mastertinner ,

I'm sorry, thank so much...

rawmind0:
Closing due to inactivity. Please reopen it, if you are still interested.

PR #174, opened 2016-06-07
add new version of gogs

mastertinner:
fix and update gogs


cloudnautique:
Where does Gogs store all of its data? Does it store anything on disk? If so, it will be lost in the upgrade.

rawmind0:
Hi @mastertinner ,

I'm sorry, thank so much...

ununseptium:
I would suggest to close this, i don't think that something will happen anymore to this.

mastertinner:
Since I deleted the source branch, I'll go ahead and close this and create a new, up to date PR.

PR #178, opened 2016-06-12
Allow custom kopf and kibana public ports

sjwoodr:
An ELK stack with elastisearch+logstash+kibana is super simple to setup in Rancher.  Yet it seems no one has done this via the catalog, *using a single rancher host*.  Elastisearch has a kopf container which binds to port 80 on the host.  Kibana sidekick also binds to port 80 on the host.  So this is obviously a problem.  

This PR is to allow for custom public ports for both kopf and kibana so that both can run on the same host.  Not ideal for production, but super useful for proof-of-concepts and demos.

### DIFF
--- a/templates/elasticsearch-2/0/rancher-compose.yml
+++ b/templates/elasticsearch-2/0/rancher-compose.yml
@@ -14,7 +14,7 @@
       description: "Unique public port for kopf"
       type: "int"
       default: 80
-      required: false
+      required: true
 elasticsearch-masters:
   metadata:
     elasticsearch:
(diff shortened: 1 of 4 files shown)


templates/elasticsearch-2/0/rancher-compose.yml
```diff
@@ -9,6 +9,12 @@
       type: "string"
       required: true
       default: "es"
+    - variable: "kopf_port"
+      label: "Public Port"
+      description: "Unique public port for kopf"
+      type: "int"
+      default: 80
+      required: false
```
cloudnautique: If they aren't required, then we would allow blank which would create a syntax error in the compose file because of the colon. I'd set it to required to avoid that.

sjwoodr: Ah, good catch.  I'll fix that.

PR #182, opened 2016-06-15
updated traefik to use new sidekick

cpoole:
Major points:

- I added the ability to set the entire traefik.toml config file through a metadata option in the rancher-compose file. This works by standing up an alpine sidekick that queries the rancher metadata api and volume mounts the traefik config into the traefik container.
- I bumped traefik to version 1.0.0-rc2
- Since I don't use host labels I removed the host affinity labels... I'm not sure what is the best way to set these since I suspect there will be many instances where people either have their own naming conventions or simply don't use them. Forcing people to tag their hosts seems odd. 
- @rawmind0 do you have any ideas as to how to make these labels dynamic? What does rancher do if a label is left empty? Could we simply have the values of the labels be catalog questions with a default empty string?


cloudnautique:
The labels could be set to soft affinity for the host. That way you can force the LBs to specific hosts if you need to. This is actually really useful if you are doing demos/POCs where you only want 1 dns entry to point to a host.

cpoole:
@cloudnautique that sounds like a great feature, unfortunately I can't get it to work. I am running 1.1.0-dev1 and if I don't have a host with the label the stack never initiates and is permanently unhealthy. However, the moment I add a label to the host the containers are scheduled. 

the dockerfile has label: ```io.rancher.scheduler.affinity:host_label_soft: traefik_lb=true```

Was this "soft" label feature added recently?

rawmind0:
@cloudnautique : totally agree with you. Only to say that, it's not only useful to do demos/POCs, it's also useful to get service working in ha, adding a round robin dns entry to the hosts ... :)

@cpoole: package was thought to be deployed with mandatory label, to take control where to expose. I choose that way, because i prefer to have that control, setting one simple label. Right now, if you don't set the label the service never will be deployed. It is the normal operation. 
I added some features in this pr https://github.com/rancher/community-catalog/pull/185 , one of them, "Configurable host label affinity in catalog". It would be ok for you? 

I think that empty string would not work, it never match. As far as i know, it's posible to put variables in your docker-compose.yml, but i think that it's not posible to put conditionals. 

That's the default scheduler policy of the package, with mandatory but configurable label:
    io.rancher.scheduler.global: 'true'
    io.rancher.scheduler.affinity:host_label: ${host_label}
    io.rancher.scheduler.affinity:container_label_ne: io.rancher.stack_service.name=$${stack_name}/$${service_name}

To disable host affinity you have to delete all the line. You need some similar to
    io.rancher.scheduler.global: 'true'
if (${host_label}){
    io.rancher.scheduler.affinity:host_label: ${host_label}
}
    io.rancher.scheduler.affinity:container_label_ne: io.rancher.stack_service.name=$${stack_name}/$${service_name}

or set host_label with the full line value...

host_label = "io.rancher.scheduler.affinity:host_label: traefik_lb=true" instead host_label = "traefik_lb=true" 

and set
    io.rancher.scheduler.global: 'true'
    ${host_label}
    io.rancher.scheduler.affinity:container_label_ne: io.rancher.stack_service.name=$${stack_name}/$${service_name}

imho it seems a little bit awful, but it's possible to do it. :)

Otherwise, your modifications changes the package completely . No one of the initial images are used. You should modify the maintainer too. :P
Actual package is integrated with rancher-metadata, that doesn't have official rancher integration yet, but it seems that you remove it. why??

I think that more than an update it's a new package. 

Best regards..

cpoole:
@rawmind0 Thanks for the response!
When I had originally removed the labels I wasnt aware of the soft affinity feature. I have since added them back in but now they are as follows:
`
io.rancher.scheduler.affinity:host_label_soft: ${host_label}
io.rancher.scheduler.affinity:container_label_ne_soft: io.rancher.stack_service.name=$${stack_name}/$${service_name}
`
This is very similar to your pull request #185 The only difference is I have added soft affinity to these two labels. 

As for the update to the image I swapped our your traefik-alpine image for the official traefik image. You have a pretty extensive dockerfile inheritance that is based on m/monit as the entry process. I'm not sure what the general consensus around the community modules are but I would think people are expecting the official image if possible. @cloudnautique maybe you can chime in? Are there any guidelines for what images Rancher wants to see in the community modules?

I'm going to go ahead and close this pull request, I have a pretty different set of requirements for how this module should work. If I need to augment any features later, I'll submit pull requests against your container files on github.

PR #161, opened 2016-05-29
volumes_form -> volumes_from typo

ashmckenzie:
When attempting to create a consul-registrator stack, I spotted this typo.

PR #183, opened 2016-06-15
Fix typo on consul data volume

chussenot:


chussenot:
Same as this one... https://github.com/rancher/community-catalog/pull/161/commits/ed670d3739300af8b14e80f2accb80fa8bb91862

cloudnautique:
@galal-hussein, can you take a look at this? I want to say if you give services the label `io.rancher.container.dns: true` you get access to Rancher DNS. That won't make the link work, but you could find service.stack at that point.

galal-hussein:
@cloudnautique It worked perfectly with enabling the DNS, and accessing consul service instead of linking and using net=host

cloudnautique:
@galal-hussein, cool. 
Should this PR be changed to use the labels? Or what is the next step?

galal-hussein:
@cloudnautique I think the PR should use the dns label instead of the link, removing net=host resulted in odd behavior though where consul started to fail.

cloudnautique:
Is this good to go now @galal-hussein

galal-hussein:
@cloudnautique Yes it works fine, tested on Docker 1.11.2 and rancher dev5

PR #185, opened 2016-06-15
Upgrade zookeeper and traefik packages

rawmind0:
Zookeeper feats:
- Upgraded to version 3.4.8,
- Decoupled config as sidekick.
- Added more params to config.
- Data volume as sidekick.
- Run service as zookeeper user.

Traefik feats:
- Upgraded to version 1.0.1
- Add ssl support. Paste key/crt in catalog.
- Run service as traefik user.
- Publish services with multidomain rules, through traefik.domain label.
-  Configurable host label affinity in catalog.
- Two modes of publish services, traefik.enable label:
  - true: service will be published as *service_name.stack_name.traefik_domain*
  - stack: service will be published as *stack_name.domain*. 

Kafka feats:
- Upgraded to version 0.10.0
- Decoupled config as sidekick.
- Added more params to config.
- Data volume as sidekick.
- Run service as kafka user.


cpoole:
@rawmind0 line 24 of the rancher compose for traefik should be https. The label says http so there are two http labels.

masone:
Thanks for the updates. I have been playing around with traefik in the last few days and stumbled upon your PR here. The `ssl_cert` is not required and falls back to `""`. It does not work for me *without* SSL cert though:

```
7/25/2016 9:38:31 AMtime="2016-07-25T07:38:31Z" level=info msg="Traefik version v1.0.0 built on 2016-07-05_11:25:54PM" 
7/25/2016 9:38:31 AMtime="2016-07-25T07:38:31Z" level=info msg="Using TOML configuration file /opt/traefik/etc/traefik.toml" 
7/25/2016 9:38:31 AMtime="2016-07-25T07:38:31Z" level=info msg="Preparing server http &{Network: Address::8080 TLS:<nil> Redirect:<nil>}" 
7/25/2016 9:38:31 AMtime="2016-07-25T07:38:31Z" level=info msg="Preparing server https &{Network: Address::8443 TLS:0xc82035bf00 Redirect:<nil>}" 
7/25/2016 9:38:31 AMtime="2016-07-25T07:38:31Z" level=fatal msg="Error creating TLS config crypto/tls: failed to find any PEM data in certificate input" 
7/25/2016 9:39:01 AM'traefik' failed to start (exit status 0) -- no output
```

rawmind0:
You were right @masone . ssl key and crt has to be required, by the moment. I set them as a required and add a default test certificate value.

Obviously, the correct way would be integrates ssl with rancher server certificates. Unfortunately, it seems that there isn't a clear way yet, isn't it??

It is fixed in the last commit, and updated traefik to v1.0.1

masone:
Thanks for making clear the SSL cert is required Raul! This makes sure noone deploys a failing server.

From my point of view, having SSL required is a step back to the previous version. I'd rather put a haproxy/letsencrypt in front of traefik. Enforcing users to handle SSL that way excludes everyone like me from using traefik.

Is there a way to run traefik without SSL?

rawmind0:
Thanks for the feedback @masone .

I push a new commit, with a new label, to be able of choose https working mode on traefik.
https_enable = <false | true | only>
  - false: Enable http enpoints and disable https ones.
  - true: Enable http and https endpoints.
  - only: Disable http endpoints and enable https ones.

Does it work for you?? :)

masone:
Wow, thanks! Just had a look [at your commit in alpine-traefik](https://github.com/rawmind0/alpine-traefik/commit/04505da458b74ad1660a39beab749e463f5c6dad) and found `TRAEFIK_ENTRYPOINTS`. It sounds like that supports the same behavior as your new `TRAEFIK_HTTPS_ENABLE` feature. How is the new one superior? I'm asking because removing the old ENV variable is probably a breaking change.

I didn't know about `TRAEFIK_ENTRYPOINTS` - if you want we can also configure a question for that instead.

rawmind0:
Good morning @masone,

with the value of TRAEFIK_HTTPS_ENABLE, it forms the adecuate TRAEFIK_ENTRYPOINTS value. It could be: 
- if TRAEFIK_HTTPS_ENABLE="false", then TRAEFIK_ENTRYPOINTS="http"
- if TRAEFIK_HTTPS_ENABLE="only", then TRAEFIK_ENTRYPOINTS="https" 
- if TRAEFIK_HTTPS_ENABLE="true", then TRAEFIK_ENTRYPOINTS="http","https"

TRAEFIK_ENTRYPOINTS variable already existed in the previous revisions. The new env variable is TRAEFIK_HTTPS_ENABLE, that works as i've explained. Additionally, i've added an option field in rancher-compose to be able to set your https working mode.

If you have any trouble, you could use this repo https://github.com/rawmind0/service-catalog where i already publish the last versions of my packages. 

Is that ok for you??  :)

masone:
Thanks for the explanation, I'll have a look at it and give the newest version in your other catalog a spin this week!

masone:
The latest update works well for me! Thanks for your help @rawmind0

PR #179, opened 2016-06-13
Dns Update (RFC2136) catalog entry

digitalLumberjack:
Add the rfc2136dns provider.

Related and depends on https://github.com/rancher/external-dns/pull/35
I set the image to rancher/external-dns:v0.4.2, thinking you will release a new version if you merge the external-dns PR.

PR #192, opened 2016-06-18
Bind9 DNS catalog entry

digitalLumberjack:
Add a Bind9 DNS in a system stack. 
Based on https://github.com/digitalLumberjack/docker-bind9
The Bind9 server is configured to serve axfr requests for the root domain, and accept updates based on RFC2136, with simple TSIG security. 
See https://github.com/rancher/external-dns/pull/35 for the dns provider that will update this server with rancher metadatas.
See https://github.com/rancher/community-catalog/pull/179 for the provider catalog entry.

The svg icons are in MIT licence, i don't know if you have a MIT licence text file in the source code that can cover the `The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.` part of the licence.


templates/bind9/0/docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+bind9:
+  image: digitallumberjack/docker-bind9:v1.0.0
+  ports: 
+   - 53:53/tcp
```
cloudnautique: I realize these are the obvious choices :) But we get a lot of community feedback to make the public ports configurable via the questions. Can you make these configurable and set the defaults to 53?

digitalLumberjack: Good point :)

PR #194, opened 2016-06-19
Upgrade Alfresco to 5.1 201605 GA

disaster37:
Upgrade Alfresco to last release and fix some issues about librairies not installed.


cloudnautique:
This PR will delete everything on the upgrade because the Postgres container will be upgraded due to the pull_image: always label on the service. 

I would recommend redoing this catalog entry in the 0 template version.  The steps needed would be:
1.) update the version of Alfresco
2.) remove pull always label from postgres
3.) create a data volume sidekick service for the PGData volume.

disaster37:
Hi, thx for your answere. Like as you say, I have updated Alfresco int the 0 template version.
I have also remove all pull always label from postgres and I have added sidekick storage container for postgres and for Alfresco.

I hope is now better ;)

templates/alfresco/0/docker-compose.yml
```diff
@@ -3,22 +3,36 @@ alfresco:
...
   links:
   - postgres:db
   stdin_open: true
   ports:
   - 8080:8080/tcp
+alfresco-data:
+  labels:
+    io.rancher.container.start_once: 'true'
+  image: alpine
+  volumes:
```
cloudnautique: Should we add a `command:` or `entrypoint:` `/bin/true`  on the data containers?

PR #196, opened 2016-06-21
Prometheus - Updated incorrect link for exporter

Rucknar:
Reported through the ticket below: 

https://github.com/infinityworksltd/Guide_Rancher_Monitoring/issues/6#issuecomment-227216325

This fixes a simple error in the linking, it wasn't picked up as the implicit DNS resolution meant it worked anyway.


templates/Prometheus/1/docker-compose.yml
```diff
@@ -67,7 +67,7 @@ grafana:
        - graf-db
     links:
       - prometheus:prometheus
-      - rancher-api-integration:rancher-api-integration
+      - prometheus-rancher-exporter:prometheus-rancher-exporter
```
jifeon: Why you changed link for `grafana`, but did not for `prometheus`?

PR #218, opened 2016-07-15
Add data container to jenkins-swarm

dpattmann:
I we mount ```/var/jenkins_home/workspace``` from the host machine the directory doesn't have the right permission. Also we don't want to write any data to the host machine.


cloudnautique:
Thanks for the PR!

 The Jenkins workspace dir is a tricky one. In some cases, if you just run inside the swarm container things are fine. If you leverage the hosts Docker with certain plugins you end up in a bit of a bind. The main problem is from the job definition, you do not know the name of the container to do a volumes_from if the data dir is in a shared volume. Since your job can end up on any of the swarm hosts.

Typically, I have relied on host bind mounts, and realize that its not a great approach. One thing that might work better, or at least fit into the container life cycle, would be a named volume that users could know ahead of time when defining jobs. Something like jenkins_workspace

Any other ideas or thoughts suggestions?

dpattmann:
I understand the problem but I think there are just some limitations you have to deal with when you are working with Jenkins, Docker and Jenkins Swarm Nodes.

A main reason for my PR is that without a sidekick data container you can't run Swarm Nodes with the Jenkins user because of the host mount directory/file permissions and you have to clean up files manually when a Swarm Node is removed.

cloudnautique:
@dpattmann, what do you think about using a named volume on the data container? Something like jenkins_home?  Sound like a reasonable compromise?

dpattmann:
@cloudnautique You mean something like this?

```
 swarm-datavolume:
   image: "busybox"
   volumes:
-    - /var/jenkins_home
+    - jenkins_home:/var/jenkins_home
   labels:
     io.rancher.container.start_once: true
   entrypoint: ["chown", "-R", "1000:1000", "/var/jenkins_home"]
```

What would happen when a second jenkins swarm container, maybe from a different stack, tries to create a volume with that name?

cloudnautique:
Yes, exactly.

 A different stack would share the same volume. You could make the volume name a user configured question. Though there would be no guarantees of uniqueness. In coming versions of Rancher this would be a non-issue, we are revamping volumes with compose v2 syntax and scoping will happen to a stack level to avoid this. 

Want to make it configurable for now?

dpattmann:
I'm not really convinced if a named volume inside a data volume container has more benefits than just a data volume container but I can upgrade my PR in the next days. ;-)

cloudnautique:
@dpattmann, look forward to the update.
The named volume is for those of us who use docker inside that will need to bind mount the workspace. Knowing the name/path is important, and knowing the container name for a volumes from is somewhat tricky because its based on Stack/Service and create index.

dpattmann:
I updated my PR

dpattmann:
I tried a lot of things but Drone can't build this PR, the `origin/pr/218` branch seems to be broken so I'm going to create a new PR.

PR #219, opened 2016-07-18
Etcd v2.3.6-4

LLParse:


templates/etcd-ha/3/README.md
```diff
@@ -1,41 +1,38 @@
 # Etcd 
 
-A distributed, highly-available key/value store written in Go.
+A distributed, highly-available key/value store for storing configuration information.
```
cloudnautique: It can store more then config info. Maybe something more like:
etcd is a distributed key value store that provides a reliable way to store data across a cluster of machines.

which is from coreOS's description.

templates/etcd-ha/3/README.md
```diff
@@ -1,41 +1,38 @@
...
 
-### Info
+### General
 
- This creates an N-node etcd cluster on top of Rancher. The bootstrap process is performed using a standalone etcd discovery node. Upon the cluster entering a running state, this discovery service will shutdown. The state of the etcd cluster is a key/value pair stored on etcd itself, in a hidden key at location `/_state`.
+This template attempts to create an N-node Etcd cluster. Only 1 node is allowed per host. If not enough hosts are available, the largest cluster possible will be built with the resources available. Adding more hosts at a later date will result in the cluster scaling up to the maximum desired size.
```
cloudnautique: 'attempts' makes it sound a bit unreliable. I get what the goal is, but maybe say something like: The template deploys a cluster with scheduling policy up to N nodes in 1 node increments ... 
Something to that end.

templates/etcd-ha/3/docker-compose.yml
```diff
@@ -1,41 +1,22 @@
...
-  - discovery
   volumes_from:
-    - data
-# containerize data volume to enable restarts and upgrades
+  - data
+
 data:
   image: busybox
-  command: /bin/true
+  entrypoint: /bin/true
```
cloudnautique: changing this will cause the data container to be destroyed on upgrade I believe. Is this needed?

templates/etcd-ha/3/rancher-compose.yml
```diff
@@ -1,22 +1,28 @@
...
+  retain_ip: true
+  scale_policy:
+    min: 1
+    max: ${SCALE}
+    increment: 1
   health_check:
     port: 2379
-    interval: 7500
+    interval: 5000
+    response_timeout: 3000
```
cloudnautique: Can we put the quorum strategy stuff in here? Or are we not using it because quorum moves depending on scale?

templates/etcd-ha/3/README.md
```diff
@@ -0,0 +1,38 @@
+# Etcd 
+
+A distributed, highly-available key/value store for storing configuration information.
+
+### General
+
+This template attempts to create an N-node Etcd cluster. Only 1 node is allowed per host. If not enough hosts are available, the largest cluster possible will be built with the resources available. Adding more hosts at a later date will result in the cluster scaling up to the maximum desired size.
+
+### Upgrades
+
```
deniseschannon: Instead of putting in limitations/what's coming in a future release within sections, could we make a whole section at the beginning or end so it' smore clear to users?

templates/etcd-ha/3/README.md
```diff
@@ -1,41 +1,38 @@
...
-Click deploy.
+If a majority of nodes are unrecoverably lost, you must re-build the cluster from one of the surviving nodes. The process involves selecting a survivor, transforming it into a standalone node, and adding new hosts so the service may scale back up to the desired size.
 
-Once the stack is deployed and assuming your application is deployed within it, you can access the etcd cluster in your application like so:
+In more detail, follow these steps:
 
-```
-etcdctl --endpoints http://etcd:2379 member list
-```
+1. Determine if your lost hosts/containers are truly unrecoverable. If a host comes back online, wait at least 5 minutes while Network Agent repairs itself and the Etcd container should return to a running state. If the hosts are truly unrecoverable, remove them from the environment.
```
aemneina: is there an easy/obvious way to determine which containers are useless vs which are candidates for recovery? it might be good to make this more clear.

templates/etcd-ha/3/README.md
```diff
@@ -1,41 +1,38 @@
...
+3. From the dropdown menu of the same container, click `Restart`. This will trigger the disaster recovery automation which will sanitize the data directory, update cluster membership to reflect a new standalone deployment, and start the node. At this point, Etcd will begin servicing requests and downstream containers should return to a functional state.
+4. In the event you experienced a majority of hosts failing simultaneously but had a surplus of hosts, you will have unhealthy Etcd containers scheduled to other hosts. Wait patiently and they will automatically join the cluster. If you did not have a surplus of hosts, add new ones and etcd will scale up to the desired size.
 
-On the etcd cluster itself, ETCDCTL_ENDPOINT environment variable is set so you may inspect like so:
+It is highly unlikely that you will ever need to perform these steps. If you've truly experienced a disaster scenario (and you didn't cause it), your infrastructure provider should be assessed for instability. If in the cloud, consider switching providers.
```
aemneina: last line not necessary? ¯\_(ツ)_/¯

templates/etcd-ha/3/docker-compose.yml
LLParse: Combining entrypoint and command allows you to specify the default executable for your image while also providing default arguments to that executable which may be overridden by the user. Let's look at an example:

```
FROM ubuntu:trusty
ENTRYPOINT ["/bin/ping","-c","3"]
CMD ["localhost"]
```

In this case, busybox image has no default entrypoint or command. Therefore, using entrypoint vs command is a matter of stylistic choice. However, if busybox had a defined entrypoint that did something which might fail or have some other side effect, we would want to override that. Otherwise, as you may have experienced, the command will be sent as arguments to the entrypoint executable, which may or may not return 0. If it fails, we lose the data volume.

I chose to make the change in an attempt to standardize. If someone were to mimic our data volume settings with an image that had an entrypoint already defined, they could accidentally lose their data. The latest four k8s templates use it as well.

cloudnautique: My question wasn't around the usage of entrypoint vs command... 

My concern is for anyone that has deployed and attempts to upgrade.. we would delete the data volume because the container definition changed and therefore needs to be upgraded.

templates/etcd-ha/3/rancher-compose.yml
LLParse: I couldn't think of a deterministic way to compute that value as etcd scales up, so for now I left it out. I think the best solution is to recompute `quorum=floor(scale/2)` each time scale changes.

PR #236, opened 2016-08-02
Removing convoy-gluster from catalog

deniseschannon:


deniseschannon:
We have removed GlusterFS and Convoy Gluster from the catalog as users were expecting a robust tool as an alternative persistent storage for Docker volumes.  However, due to lack of active maintenance, we cannot recommend this solution going forward.

Instead, we recommend and certify Convoy NFS, which is actively maintained by Rancher.

PR #242, opened 2016-08-06
[TASK] change taiga templates

KartoffelToby:
Change the Docker Images, working Conf


lilnate22:
+1 current taiga doesnt work at all

cloudnautique:
@KartoffelToby,

 Thanks for taking a pass at this. I see there are some python and json files in the folder. Those will not be picked up by Rancher, and need to be embedded in a container. 

 Also, while we are updating the template, we should add a data volume side kick to the Postgres container. That way we can upgrade it and make scheduling decisions about it later.

KartoffelToby:
@cloudnautique OK. I'll copy the files to the Image and add the Volume.

cloudnautique:
Ping @KartoffelToby

KartoffelToby:
Sorry i was really busy.

Config files are now in the Container.
The postgres Container has a Volume that can be set via rancher-compose file

lilnate22:
+1 @KartoffelToby 👍

MichelDiz:
This is not working. I cannot access either locally, or even in the network. (using a custom host locally) There's nothing in "TAIGA_HOSTNAME taiga.test.de"

PR #247, opened 2016-08-18
Added named data container to jenkins-swarm

dpattmann:
See https://github.com/rancher/community-catalog/pull/218 for more informations

PR #249, opened 2016-08-19
Drone Build: Set right git branch for rancher-catalog-service

dpattmann:
I created two PRs, https://github.com/rancher/community-catalog/pull/247 and https://github.com/rancher/community-catalog/pull/218, and on both the Drone builds are failing. I have debugged the build process and found the problem.

The `rancher-catalog-service` updates its catalog during the first request from the master branch of the git repository but the new templates are in a PR branch.

I didn't find a way the set the git branch through the command line so I created a config file and update the branch name if it is not **master**.


cloudnautique:
Cool, Thank you for digging into this!

PR #254, opened 2016-08-24
Adding TeamCity Stack to the Rancher

matisku:
**What is inside TeamCity Stack?**
-     TeamCity Server
-     Postgres Database
-     Scalable TeamCity Agents

**Instructions**
TeamCity Stack will create "teamcity" postgres database with teamcity user.
I included additional variables for `http_proxy` and `https_proxy`, which can be useful in corporate environment. Once TeamCity will start, make sure you setup the same information in setup page.

**TeamCity Agents**
TeamCity Agents will start automatically and connect to the TeamCity Server.
Agents should be available in TeamCity Server in about 5 minutes after stack's start.


cloudnautique:
Im ok with the teamcity side of things, but why is gogs being upgraded in this PR? Can it be split out?

templates/gogs/config.yml
```diff
@@ -1,5 +1,4 @@
 name: Gogs
 description: |
     A painless self-hosted Git service
-version: v0.8.43
-category: Git
+version: v0.8.43
```
cloudnautique: Why remove the category?

cloudnautique:
Looks like there is one last "gogs" thing in the mix. We can merge once thats removed.

PR #262, opened 2016-08-30
smart(?)er wait for listening catalog service

nrvale0:
We were getting test flakes in CI because the static wait for start of Rancher Catalog service was expiring without service start. Make the wait check for listening tcp/8088 as a reasonable predicate for working Catalog service.

@LLParse @cloudnautique


LLParse:
By using `netstat`, we introduce a dependency...which should be OK as long as tests are always performed from within a Linux container with the 'net-tools' package installed. For reference, the [/dev/tcp](http://www.tldp.org/LDP/abs/html/devref1.html#DEVTCP) device file can be used in bash scripts to do the same thing in a platform-agnostic way.

LGTM

nrvale0:
@LLParse @cloudnautique Had some problems getting the fdes sorted out with /dev/tcp so falling back to a curl. I can't imagine we'd ever have a build container which doesn't have curl|wget.

LLParse:
If you're going to use curl, could you provide the `-s` flag so we don't get ugly output in the logs:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--    
```

nrvale0:
@LLParse With the exception of a little stderr output on failed poll as progress/feedback, that's less noisy.

cloudnautique:
@nrvale0 can you rebase?

nrvale0:
@cloudnautique Done

rawmind0:
Hi @nrvale0 ,

I'm sorry, thank so much...

PR #263, opened 2016-08-31
sysdig-cloud: Cut logs at 100mb and make a system service

jmreicha:
Adds `log-opt` to cut logs every 100mb so disk sizes don't grow and makes it a system service.


jmreicha:
Do I need to worry about the failing checks?

cloudnautique:
Would you mind rebasing with master just to make sure. We had some issues when we cut over CI systems.

jmreicha:
@cloudnautique Done.

cloudnautique:
@jmreicha, looks like this was a merge commit. 

We need it rebased, I don't think I can merge this with out it getting wonky.

PR #269, opened 2016-09-07
feat(kubernetes): Traefik load balancer

germanramos:
Traefik load balancer for Rancher with Kubernetes.

Provide load balanced access to all services exposed with an Ingress rule.

Fully configurable and extremely useful.


rawmind0:
LGTM....

Thanks @germanramos for you PR and contribute.... :)

PR #275, opened 2016-09-09
Fixes versions in templates

mrajashree:
Some templates in community catalogs do not have versions in the correct format. They are used by catalog-service, and the library semver being used by catalog service to compare versions requires all versions to be of the format: `X.Y.Z` or `X.Y.Z-alpha` etc, where X, Y and Z have to be integers.

### DIFF
--- a/swarm-templates/wordpress/0/rancher-compose.yml
+++ b/swarm-templates/wordpress/0/rancher-compose.yml
@@ -1,6 +1,6 @@
 .catalog:
   name: "Wordpress"
-  version: latest-rancher1
+  version: 1.0.0-latest-rancher1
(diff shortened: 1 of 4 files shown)


cloudnautique:
Can you rebase please? I want to make sure the CI run is green.

mrajashree:
Yes even I was looking at why tests failed. I'll rebase and push the changes again

mrajashree:
@cloudnautique Can you please review this?

cloudnautique:
LGTM, should we be adding a test that validates versioning? /cc @vincent99

vincent99:
Yes we probably should have a test that checks that `blang/semver` can parse the version.

PR #301, opened 2016-10-01
Add a Confluence template

SanderKleykens:
This pull request adds a Confluence template to the catalog


cloudnautique:
This failed CI, can you take a look? It looks like the version you have in config.yml and rancher-compose.yml are different. A lot of times we follow the convention where its "<software_version>-<buildnumber>" or something to that effect.

SanderKleykens:
Hi @cloudnautique, I didn't realize the version in config.yml were supposed to be the same as the version in (one of) the rancher-compose.yml files. I learned something new!
Anyways, I fixed the issue and the build is successful now. Thanks for your insight!

PR #304, opened 2016-10-05
sysdig-cloud k8s

alexsysdig:
Container-Native Application and Infrastructure Monitoring


galal-hussein:
The template should use Questions defined in rancher-compose file as variables in the YAML file to define properties of the Daemonset:

```
        - name: ACCESS_KEY                                  
          value: ${SDC_ACCESS_KEY}
        - name: TAGS                                    
          value: ${SDC_TAGS} 
        - name: ADDITIONAL_CONF                        
          value: ${SDC_ADDITIONAL_CONF}
```

Also you should add the ${VERSION} to the image of sysdig cloud agent:
```
image: sysdig/agent:${VERSION}
```
And as for HOST_EXCLUDE_LABEL, you can use something like node affinity to not schedule sysdig pods on a node with certain label:

http://kubernetes.io/docs/user-guide/node-selection/

rawmind0:
Adding some comments to @galal-hussein ones.....

/lib/modules has to be mounted as read-write to be able to build and save sysdig kernel modules...

```
        - mountPath: /host/lib/modules
          name: modules-vol
          readOnly: true
```

alexsysdig:
SDC_ACCESS_KEY, SDC_TAGS and SDC_ADDITIONAL_CONF are defined int the rancher-compose.yml and they generate questions. Is there something else you want me to add?

alexsysdig:
Regarding the image: sysdig/agent:${VERSION}. We would like to have our clients pulling always the latest. Is that ok if I change to image: sysdig/agent:latest?

alexsysdig:
Updated files for PR comments

03cranec:
sorry folks, had some github confusion over here. I've opened a new PR here: https://github.com/rancher/community-catalog/pull/318

i think this addresses your comments on the original PR here.

let me and @alexsysdig know if we need to do anything else here. we've got a rancher+sysdig meet up this week so it would be great to get this live asap. happy to hop on a quick call tomorrow if needed. thanks all!

PR #305, opened 2016-10-06
Update Prometheus Stack

dominikhahn:
- Update images
- Make public port expose flexible
- Add option to deploy the stack on specific "allowed" hosts


galal-hussein:
@dominikhahn Looks like the fork is removed, can you make the PR available in another repository.

PR #309, opened 2016-10-12
ELB Classic LB template

janeczku:
Depends on `rancher/external-lb:v0.2.1`


galal-hussein:
Tested and worked on Rancher master image.

LGTM

PR #319, opened 2016-10-26
Adding QuasarDB Community database catalog entry 2.0.0-rc8

AlexisDucastel:
Hi everybody,

I would like to offer this catalog entry for QuasarDB community. It is a key-value store written in C++ which is very light and very efficient. With only two nodes of commodity hardware we can get up to 450k req/s on a simple gigabit network.

For more information on QuasarDB see quasardb.net.


AlexisDucastel:
cc @bblanchon @edouarda from QuasarDB

PR #333, opened 2016-11-20
Dns Updates

deniseschannon:
* move external DNS templates to infra-templates
* Add min/max versions to external DNS templates
* Update category to External DNS instead of Rancher services


vincent99:
LGTM

Issue #239, opened 2016-08-04
Gitlab Template doesn't start


galal-hussein:
Rancher Version: v1.2.0-pre1

The Gitlab catalog entry doesn't start and the **gitlab_gitlab-server_1** container stopped without throwing any error in the logs.

cdrx:
On a related note, the version of GitLab used in the template (8.5) is quite old now, current version is 8.11.

galal-hussein:
The Issue was conflict between host and container on port 22

PR #335, opened 2016-11-22
Add more port flexibility to gitlab

ununseptium:
The Gitlab catalog will fail basically all the time if you try to add it to a host which already has for example port 22 in use. (Which is very likely) So i added the option to change the ports which are going to be exposed to something that fits your environment. 

Please be aware that changing port 22 to something else _could_ cause unknown exceptions.

Read more about that topic: 
 - https://major.io/2013/05/14/changing-your-ssh-servers-port-from-the-default-is-it-worth-it/ 
 - https://www.adayinthelifeof.nl/2012/03/12/why-putting-ssh-on-another-port-than-22-is-bad-idea/


galal-hussein:
Thanks for the PR, this will solve this issue https://github.com/rancher/community-catalog/issues/239

galal-hussein:
LGTM

PR #337, opened 2016-11-24
Update config.yml

faddat:
Update version number.


faddat:
This didn't build because it wasn't together with my other PR's.  Should have made them as a group.

PR #344, opened 2016-11-28
Mesos slave cni network model changes, parameterize mesos-master/zookeeper scale

LLParse:
More specifically, this makes the mesos-slave containers' overlay network IP resolvable by the container's hostname. This allows us to run Mesos frameworks such as Kafka which make use of the Mesos containerizer (process lives inside of mesos-slave container).


galal-hussein:
LGTM

PR #346, opened 2016-11-30
remove swarmkit

LLParse:
moving to rancher-catalog


deniseschannon:
This should remain in community-catalog.

PR #222, opened 2016-07-20
Update Jenkins catalog entry to latest (2.7.1)

ovidiub13:
I've just bumped the version number. I'd like to also check if the https://hub.docker.com/r/rancher/jenkins-plugins/ image is still compatible with this version, but the sources are nowhere to be found.

PR #349, opened 2016-12-01
Update jenkins

ovidiub13:
This is a redone of #222.


ovidiub13:
I've updated to the current latest: 2.19.4, which I'm currently using, but not from the catalog.

rawmind0:
Hi @ovidiub13 ... 

I've tested the package and seems to work well doing a new installation and updating from previous one. 
LGTM

Thanks so much for your patience and your effort...

ovidiub13:
Thank you for the approval, but I can't merge it. I don't have the permissions.

rawmind0:
Don't worry...I merge it right now... :)

Thanks so much for the PR and the contribution....

PR #352, opened 2016-12-05
Adding SonarQube Stack

matisku:


templates/sonarqube/0/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
+sonarqube-data:
+    image: busybox
+    tty: true
```
galal-hussein: @matisku can you add `net: none` to the data containers, also there is no need for them to stay running in the background, you may want to use `start_once` label, thanks.

matisku: Thanks for a tip!. Done.

galal-hussein:
@matisku Thanks, LGTM

PR #356, opened 2016-12-06
Adding Elasticsearch+fluentd+kibana template

renhuiyang:
EFK service can help you store/query logging messages in your applications. All you have to do is tell your docker containers to forward log messages to running fluend (by using --log-driver=fluentd).


rawmind0:
Hi @renhuiyang .. Thanks so much about you PR and contribute... 

Have you consider to use the already created elasticsearch and kibana packages?? You could add only fluentd and reuse the already created packages. What do you think?? :)

cloudnautique:
I would be interested in seeing a stand alone fluentd service.

cjellick:

PR #358, opened 2016-12-09
Added HackMD stack

elboletaire:
Added template for [HackMD](https://github.com/hackmdio/hackmd)


templates/hackmd/0/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
+hackmdPostgres:
+  image: postgres
+  environment:
+    POSTGRES_USER: hackmd
```
galal-hussein: @elboletaire The postgres connection information should be also variables.

templates/hackmd/0/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+hackmd:
+  image: hackmdio/hackmd:0.4.5
+  environment:
+    POSTGRES_USER: hackmd
+    POSTGRES_PASSWORD: hackmdpass
+    NODE_ENV: "${NODE_ENV}"
+    DEBUG: "${DEBUG}"
+    HMD_DOMAIN: "${HMD_DOMAIN}"
+    HMD_URL_PATH: "${HMD_URL_PATH}"
+    HMD_PORT: "${HMD_PORT}"
```
galal-hussein: @elboletaire Can you export this port or create a loadbalancer service to serve the web app for hackmd, thanks.

templates/hackmd/0/docker-compose.yml
elboletaire: Ok I'll change it.

templates/hackmd/0/docker-compose.yml
galal-hussein: as an example of hackmd lb:

in docker-compose
```
hackmdlb:
    image: rancher/lb-service-haproxy:v0.4.6
    ports:
    - 8000:8000/tcp
    labels:
      io.rancher.container.agent.role: environmentAdmin
      io.rancher.container.create_agent: 'true'
```
in rancher-compose
```
version: '2'
services:
  hackmdlb:
    scale: 1
    start_on_create: true
    lb_config:
      certs: []
      port_rules:
      - path: ''
        priority: 1
        protocol: http
        service: hackmd
        source_port: 8000
        target_port: 8000
    health_check:
      response_timeout: 2000
      healthy_threshold: 2
      port: 42
      unhealthy_threshold: 3
      interval: 2000
```
This example assumes that you set HMD_PORT to 8000, you can learn more about adding load balancer service through docker-compose from the [official documentation](https://docs.rancher.com/rancher/v1.2/en/cattle/adding-load-balancers/#adding-a-load-balancer-with-rancher-compose).

elboletaire: As I tould you previously in other MR, I actually have the rancher load-balancer enabled and that's why I don't like any rancher app to have ports open, because there's no need to. This way you just need to point to that container with rancher's lb and set which internal port it works on.

I needed to edit each template I've used from this repo because most of them are node apps working at port 3000, making them all collide.

If you want I'll do like in redmine stack, where I've not opened the port but I've added a readme file specifying it; but if I can prevent adding that rule, I would.

galal-hussein: @elboletaire If you don't want to expose a node port for the template then you can add a README file explaining how to use the template and in it you can explain that the user can open a port manually for the container or add a separate load balancer.

elboletaire: I'll do that then 😃 

PS. After these days of "holidays", OC.

elboletaire:
If this is not gonna be merged, close it. I'm not using rancher anymore, and thus, I'm not spending more time in this catalogs.

rawmind0:
@elboletaire , thanks for the advise....

PR #359, opened 2016-12-09
Added emby stack

elboletaire:
Added [emby](https://github.com/MediaBrowser/Emby) stack


templates/emby/0/docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+emby:
+  image: emby/embyserver:latest
+  net: host
```
galal-hussein: @elboletaire Is there a specific need to use `net: host` ?

elboletaire: It was specified by the old documentation. The new way is a bit different and I'm not sure what it implicates talking about networking.

With the old system still works, but without the `net: host` probably won't. Emby needs a lot of ports due to the amount of services it has.

elboletaire: Also, if you  check the [current docker installation instructions][instructions] you can see that after emby is installed you need to run a command in the machine host to start emby. 

I'm not sure at all, but I think that's not what we expect when we create a rancher stack (IMHO).

I've been installing emby with docker for a while and the old way works like a charm.

[instructions]: https://emby.media/linux-server.html

elboletaire:
If this is not gonna be merged, close it. I'm not using rancher anymore, and thus, I'm not spending more time in this catalogs.

rawmind0:
@elboletaire , thanks for the advise....

elboletaire:
No es nada personal Raúl 😉

rawmind0:
No te preocupes Óscar, no me lo he tomado como nada personal... :)

elboletaire:
Al ver los puntos suspensivos en ambos issues parecía que había ofendido o algo... no era la intención.

Gracias por el curro que os pegáis, que no utilice rancher no quiere decir que no valore la faena que hacéis así como el producto que estáis desarrollando 😉

rawmind0:
Disculpa el malentendido...Suelo poner puntos suspensivos, pero sin ninguna animosidad... 

Muchas gracias por tus palabras y tu apoyo... :)

PR #361, opened 2016-12-09
Added netdata stack

elboletaire:
Added [netdata](https://github.com/firehol/netdata) stack


galal-hussein:
@elboletaire Thanks for the contribution, you might want to use `  io.rancher.scheduler.global: 'true'` to install copy of the container on each server, also to expose the dashboard of netdata you will need to export the NETDATA_PORT

elboletaire:
Yeah but if you use the load-balancer there's no need to export them. That's why I prefer not to expose ports on rancher stacks.

elboletaire:
> ... you might want to use io.rancher.scheduler.global: 'true' to install copy of the container on each server

Didn't know about this feature. I'll add that tag then.

galal-hussein:
@elboletaire Thanks a lot, this will be a great addition to the monitoring application.

LGTM.

PR #363, opened 2016-12-12
make distinct templates for different versions

demarant:
I created two distinct templates for Plone 4.3 and Plone 5.0 instead of being a question in the rancher-compose. This is more in line with rancher templates versions and gives the ability to upgrade properly between versions. Already discussed and reviewed by maintainer @avoinea. I also made some pointer in the READMEs to the plone documentation site.


galal-hussein:
@demarant Thanks for the contribution, I reviewed and tested your contribution and everything seems to be working successfully. 

LGTM

PR #364, opened 2016-12-12
Added dokuwiki stack

ununseptium:
I tested it on my site and it seems to work. Please report any issues here.

If you encounter weird css behaviour: I've reported this issue to the dokuwiki people and they're aware of it. It's an php related issue and a valid hotfix for this would be to change the line `169` inside the file `/var/www/html/lib/exe/css.php` from 

```
$less->importDir[] = DOKU_INC;
```

to 

```
$less->importDir = DOKU_INC;
```

I also included this hotfix in the latest version of the dockerfile so there should be no need to change anything.


galal-hussein:
@ununseptium Thanks for your contribution, the image appears to be broken, also when deploying the stack, i couldn't find Admin option to add/remove users, i just want to make sure the deployment was correct.

ununseptium:
- I think the image should work now. 
- I also added more informations about the setup process as well.
- The user management part can be found if you navigate to: `doku.php?id=start&do=admin&page=usermanager`

It looks like this:
bildschirmfoto 2016-12-16 um 11 36 53

galal-hussein:
@ununseptium I can see the image now and initialize the wiki, one last note before merging, i couldn't connect to the wiki via HTTPS connection, however you export 443.

ununseptium:
I have to investigate this later. For now the template should be fine with port 80 exposed only. I'll add SSL/TLS whenever i'll update the underlying dockerfile.

galal-hussein:
@ununseptium Thanks, LGTM!

PR #368, opened 2016-12-14
Add redmine stack

elboletaire:
Added [redmine][redmine] stack

[redmine]: http://www.redmine.org


elboletaire:
I have a question about rancher and versions, now that I'm making this PR.

Is there any way to have alternate branch versions? I mean... in this case there's a Redmine Dockerfile that uses Passenger instead of WEBrick as a web server. 

Can I define any kind of label or something to separate both versions? Or should I create two different templates?

ununseptium:
Yes there's a way. You may already wondered why the folders where you put the docker and rancher compose files are called '0', '1', '2' and so on... you can put different versions into different folders. 

Just take a look at the Cloudflare Template for example:

bildschirmfoto 2016-12-14 um 07 08 45
Here you see the folder structure

bildschirmfoto 2016-12-14 um 07 09 37
and here you see the corresponding versions inside rancher

elboletaire:
Well I have that on this redmine stack actually, but if you choose the first version rancher always tells you there's an update for it, due to the next folders.

> By providing a new version folder number, it provides a way to upgrade your stack from a previous version of the template. Alternatively, you could update the templates in the 0 folder and just re-deploy the entry.

And reading this, I see that's the normal way. Then I don't understand why the "upgrade" buttons appear in rancher when there are folders next to the current version, if this folders are for different tags they should not show the upgrade button unless there's actually a new version for the current tag.

ununseptium:
That's a good point i completely understand. Currently Rancher just checks if a "newer" version of a template is available. 

This could be either 
 - a completely new template (new / higher folder number)
 - a change to the currently used template

The whole tag approach doesn't quite fit in the current concept of how rancher checks for newer versions. 

I think rancher shouldn't check for newer versions via this "higher folder number" thingy or it should differentiate between "updated template version (current template version e.g. 0 updated)" and "new template version (there's a new template avaiable for this software e.g. 0 in use and a 1 or higher number available". Something like this, you get the point. (Just an example).

But yeah i think since Docker supports the use of different tags, this should be something that Rancer consideres while thinking about updating templates.

galal-hussein:
@elboletaire Thanks so much for the PR, as for the versions issue, you can add versions like 3.3-passenger, you might want to look into the `upgrade_from` where you can add a specific version to upgrade from, for example if you want 3.2 to upgrade to 3.3 and not 3.1 to upgrade to 3.3 then you can add the following in the 3.3 template rancher-compose.yml:

`upgrade_from: 3.2.0`

That way only 3.2 can upgrade to 3.3 but 3.1 can't upgrade to 3.3, you can also add `upgrade_from: <3.3.0` to the 3.3.0-passenger version to force the 3.3.0 version not to upgrade to the passenger version.

elboletaire:
Oh! Thanks @galal-hussein let me change some things before merging this then 😃

rawmind0:
Superseded by https://github.com/rancher/community-catalog/pull/463 ...

elboletaire:
Thanks @rawmind0, please see both #359 and #358 to also take actions on them if necessary.

PR #369, opened 2016-12-15
Added cloud9 templates for catlle and k8s

rawmind0:
Added catalog templates to deploy cloud9 sdk, to share a dev environment.

It could be auto published throught traefik in cattle and ingress in k8s.


galal-hussein:
Reviewed and tested on Rancher v1.2.0, LGTM.

PR #370, opened 2016-12-16
added drupal to community-catalog

slimatic:


templates/drupal/0/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: '2'
+services:
+  web:
+    image: drupal:${drupal_tag}
```
galal-hussein: @slimatic Thanks for your contribution to the community catalog, the template reviewed and tested, the only note is that you can use the [versioning system](http://docs.rancher.com/rancher/v1.2/en/catalog/private-catalog/#directory-structure) for creating custom catalogs instead of using a variable to specify the version for drupal and mysql,

slimatic: @galal-hussein Thanks for the feedback. I can certainly make those changes however, creating custom catalogs, in my opinion, does not appear to be the best use of the versioning system for both drupal and mysql since the configurations could be many. The variable option allows a user to say I want to use drupal image version X with mysql Y to match the users development or production environment. Using the template will be unpractical.

slimatic:
@galal-hussein I have made the changes per your recommendation. Please let me know if you have any further questions or suggestions. Thanks.

slimatic:
@galal-hussein I never heard back regarding this pull request.

rawmind0:
Hi @slimatic ... thanks so much for the PR and sorry for the delay...

LGTM...

PR #372, opened 2016-12-16
added git-proxy service for convenient access to private repos

myfoxtail:
Provide an easy access to private repos for automation services through a git proxy container. Connect private cattle and install services from private repos.

### DIFF
--- a/templates/git-proxy/0/README.md
+++ b/templates/git-proxy/0/README.md
@@ -1,0 +1,10 @@
+# Git Proxy
+
+Easy access to private repos for automation services.
+
+Git Proxy exists to organize a painful and secure access to private repos for automation services.
+
+## Installation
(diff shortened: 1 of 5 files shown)


galal-hussein:
@myfoxtail Thanks for your contribution, if the service is intended to configure the **rancher server** to use private catalogs then it won't work because there is no way for the rancher server to connect to the `proxy.git-proxy` service, i think that's why you manipulate the resolv.conf which still doesn't configure the rancher service to see the services.

myfoxtail:
Hi @galal-hussein, thank you. Yeah... I thought about other solutions but still have no idea how to make it without changing something at the rancher server. Anyway that would be really great if rancher could allow to see services from other services like ```proxy.git-proxy```.

cjellick:

PR #378, opened 2016-12-21
Update traefik to v1.1.2 and some bugfixes in v1.1.1

rawmind0:
Added feature traefik.alias service label to publish it with alternate name.


galal-hussein:
LGTM

PR #383, opened 2016-12-25
Added options for GoCD Server and Agent v16.12.0

richbos:
Updated to allow selection of GoCD Server and Agent v16.12.0.

Deployment of GoCD Server with GoCD Agent auto-reg tested in Rancher 1.2.2

Incorrect agent image pull corrected after re-testing in commit b33fe59.


richbos:
PR withdrawn. Re-visit required to update agent image with Docker as per http://rancher.com/deploying-gocd-using-docker-compose-on-rancher/.

PR #399, opened 2017-01-12
Implement Plugin inclusion

rockaut:


rawmind0:
Hi @rockaut .. Thanks so much for the PR and contribute....

Very cool upgrade with plugin inclusion... :)

But I think it's not a good idea to have the package version tagged as latest..docker tag version neither...Because you loose the control about your version, update features of the catalog and is not easy to reproduce.... 

Sorry about that, but could you please do the opposite with versions?? I mean, instead of set template version and docker image tag as latest, could you set them as an specific version, for example 4.1.1 (i think this is the latest right now)??

thanks...

rockaut:
Hmm... i see what you mean. Will push another one.

rawmind0:
Thanks for the changes @rockaut ... :)

One more thing, could you also change docker label at docker-compose file, please??

Line 2...
```
--  image: grafana/grafana:latest
++  image: grafana/grafana:4.1.1
```

Thanks...

rawmind0:
Thanks @rockaut ... 

LGTM

PR #401, opened 2017-01-13
Add kibana v4.6.3 package

rwanyoike:
Follow the latest v4.x.


galal-hussein:
@rwanyoike Thanks for your contribution!

LGTM

mattkenn4545:
This version is incompatible with the included elasticsearch 2 stack in the catalog.

galal-hussein:
@mattkenn4545 You can still choose the older version of kibana that works with the current Elasticsearch in the stack.

PR #405, opened 2017-01-20
Sentry:Latest

### LINKED ISSUE #404: Sentry Catalog
Hi,
   It seems like sentry_email_password has been duplicated with types: password & string.

https://github.com/rancher/community-catalog/blob/master/templates/sentry/0/rancher-compose.yml#L49

https://github.com/rancher/community-catalog/blob/master/templates/sentry/0/rancher-compose.yml#L55

Also, will it  be possible to update this to latest version?

Thanks

wahmedswl:
Hi,
   Sentry along with all its dependent images have been updated to use latest version. Also, it will fix #404 where ```sentry_email_password``` was duplicated.

Thanks


galal-hussein:
@wahmedswl Can you also add a fix to the duplicate issue in the previous version, also it will be more accurate if you changed latest with the version number.

PR #406, opened 2017-01-25
Swarm public network support (Docker 1.13)

LLParse:
Not ready for merge


vincent99:
LGTM

infra-templates/swarm/3/rancher-compose.yml
```diff
@@ -0,0 +1,24 @@
+.catalog:
+  version: v1.13-beta.1
+  minimum_rancher_version: 1.2.2
```
deniseschannon: Would we want to change the Rancher version to 1.4.0 or something later as Docker 1.13 is currently "unsupported"?

LLParse:
Well, we also don't officially support Swarm (hence community-catalog), so maybe it's okay in this case? I've tested with Docker 1.13 and it seems to work well. I'm open either way

LLParse:
Per discussion this AM, Docker 1.13 doesn't work in some situations that aren't immediately apparent (but I've recently run into). So yes, let's hold off on merging this until there is a Rancher version with support.

PR #413, opened 2017-02-01
Bump Weave Scope to version 1.1.0

errordeveloper:


errordeveloper:
cc @faddat

galal-hussein:
@errordeveloper Thanks for your contribution!

Can the new version be added as a separate folder `1/`, so that the user can have the ability to select version to deploy.

errordeveloper:
> Can the new version be added as a separate folder `1/`...

I was wondering whether I had to do this or not, but previously `weavescope` was updated in `0/`...

> ...so that the user can have the ability to select version to deploy.

From our perspective, there is no good reason to not use latest version available.

galal-hussein:
@errordeveloper This will depend on the application, so my question is weavescope newer version will have major distinct features than the previous version that will make sense to separate the two versions?

errordeveloper:
> ...so my question is weavescope newer version will have major distinct features than the previous version that will make sense to separate the two versions?

I don't think so. Please see [release notes](https://github.com/weaveworks/scope/blob/v1.1.0/CHANGELOG.md) for details.

galal-hussein:
@errordeveloper Thanks for the clarification.

LGTM

errordeveloper:
Thanks!!

PR #415, opened 2017-02-02
Add volume plugin tiles for NetApp storage platforms to the infra catalog

innergy:
Configure and deploy 4 different flavors of the Rancher-enhanced NetApp Docker Volume Plugin (nDVP) directly from the infrastructure portion of the Rancher catalog.


innergy:
Going to migrate these to a different location based on a discussion with Will.

PR #421, opened 2017-02-06
feat(): add template for etherpad

oupala:
This pull request adds a template for etherpad, the real-time online collaboration platform.


oupala:
Should I change something on the template so that MariaDB data are not stored on the MariaDB container?

I haven't found any *best practice* about that. Have I missed something?

rawmind0:
Hi @oupala .. thanks for your cool PR and contribute..

Have you consider to use the galera cluster package instead add the database inseide the package??  You could deploy it and link your service with it, and take profit of the already created packaged. What do you think??

Best regards....

templates/etherpad/0/rancher-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    - variable: etherpad_db_password
+      description: "etherpad password of the etherpad database"
+      label: "Database password"
+      required: true
+      type: "password"
+    - variable: etherpad_db_root_password
+      description: "root password of the etherpad database"
+      label: "Database root password"
+      required: true
+      type: "password"
```
rawmind0: # Add service combo to select database service. You need to deploy it before.
    - variable: "etherpad_db_service"
      description: |
        Choose the database service to use. 
        It's really recommended a dedicated database service just for Etherpad.
      label: "Database stack/service"
      default: "galera/galera-lb"
      required: true
      type: "service"

templates/etherpad/0/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+    DB_TYPE: mysql
+    DB_PORT: 3306
+    DB_USER: etherpad-db-user
+    DB_PASS: ${etherpad_db_password}
+    DB_NAME: etherpad-db-name
+  links:
+  - etherpad-db:mysql
+  ports:
+  - ${public_port}:9001/tcp
+
```
rawmind0: # Remove database definition. Re-use the galera cluster catalog package.

templates/etherpad/0/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+    FAVICON_URL: ${etherpad_favicon}
+    ADMIN_PASS: ${etherpad_admin_password}
+    DB_HOST: etherpad-db
+    DB_TYPE: mysql
+    DB_PORT: 3306
+    DB_USER: etherpad-db-user
+    DB_PASS: ${etherpad_db_password}
+    DB_NAME: etherpad-db-name
+  links:
+  - etherpad-db:mysql
```
rawmind0: # Set the link to database service
--  - etherpad-db:mysql
++  - ${etherpad_db_service}:mysql

rawmind0 (changes requested):

oupala:
I tried to instantiate a galera cluster for testing purposes, but I had an error:

>  (Expected state running but got error: Error response from daemon: oci runtime error: process_linux.go:330: running prestart hook 0 caused "fork/exec /usr/bin/dockerd (deleted): no such file or directory: ")

Could it be because galera cluster requires a lot of memory (which I don't have on my machine)?

Etherpad should be usable on a standard low-power server.

rawmind0:
Mmmm...It's so strange...I've tested galera cluster this morning and it was working as expected.. How are the server you are trying to run?? 

It think it could be a good idea to launch etherpad in a cluster db.. :)

oupala:
I wrote another catalog where etherpad is using the galera cluster. Unfortunately I can't get galera to work on my test machine so I can't test my new etherpad catalog item.

In fact, I can't run the galera cluster on its own. Could it be because of lack of memory? My test machine has 4 Gio of ram.

If anyone can help me, I'd be glad to add the galera storage to my etherpad catalog item.

oupala:
I finally succeeded to instantiate a galera cluster.

When I try to instantiate a new template of etherpad (using galera cluster as a service), I set `Database stack/service` to use the `galera-lb` I've created previously. I then get the following error: `Error (Service 'etherpad' has a link to service 'galera/galera-lb' which is undefined)`. I can't go further as I don't understand why I dot have this error.

oupala:
And just to clarify, using a galera cluster instead of a simple MariaDB instance means that you switch from a one step creation:

1. create an etherpad service that integrates a MariaDB service

to a two step creation:

1. create a galera cluster instance
2. create an etherpad service that uses the previously created galera cluster

If I'm right, it would be a shame that the galera cluster can't be declared *inside* the etherpad service.

oupala:
@all I need help to understand how to use galera cluster. Otherwise, I won't be able to improve my pull request enough to get it merged. Thanks.

rawmind0:
Hi @oupala..you are right. Using galera cluster, would be a two step creation:

- create a galera cluster instance
- create an etherpad service that uses the previously created galera cluster. Service field in rancjer compose.

May be it would be a shame, but imho i think is better to maintain, as galera-cluster has its own maintainer. :)

Could you please go deeper with details about you previous service link issues??

oupala:
Thanks @rawmind0 for answering me. I feel now less lonely!

It is great to keep galera-cluster in its own repository. But I thought rancher could create the galera cluster if mentionned as a dependency of another project. This way we could instantiate an app (etherpad for example) in one step, having a robust data storage.

Do you think it could be possible to add this feature to rancher ?

My link issue is that I can't succeed to use the newly created galera-cluster from the etherpad instance I'm trying to create. When I try to use the galera-cluster from the etherpad instance, I get an error. See [this comment)(/rancher/community-catalog/pull/421#issuecomment-289791754) for more informations.

rawmind0:
I agree with you about it would be nice the posibility to define dependencies in catalog package, but unfortunately we don't have it, but the moment... :(

When you define a service field in catalog (view first changes request that i did)...It generates a selection combo of existing services...Are you selecting your galera service from the selection combo or your are typing it in a text box??

oupala:
Thanks @rawmind0 for your answer. I defined a service in the catalog exactly the way you requested it in the first change request. I have a selection combo box of existing services when I try to create an etherpad instance.

See this screenshot:

screenshot

oupala:
As I get no answer from the community. I think I will soon give up and cancel this pull request.

Too bad for etherpad and for the rancher community.

cjellick:
Sorry this never made it across the finish line. Thanks for all the work you put into it. Closing.

PR #423, opened 2017-02-07
Change teamcity category to Continuous Integration

galal-hussein:


rawmind0:
Thanks @galal-hussein ...

LGTM

PR #427, opened 2017-02-09
Update Portworx to support CoreOS

jsilberm:


jsilberm:
Ping.

Any issues?  

If not, then merge please?

cloudnautique:
This will not show up as an upgrade for users. It would only take effect for new installations. Is that the desired outcome? 

If not, a `1` folder should be created with the new files, and the config yaml updated with the new version. That way users would know to update.

There is also a risk using a non-versioned tag that is updated in a streaming fashion. What is `edge` today might not be `edge` tomorrow when a new host is added. And the user isn't aware of any difference.

jsilberm:
Understood and intended.     Most customers only use px-dev for trial/PoC, and then move on to our commercially supported version.   So having this only take effect for new installations is just fine.

And it's same story with using "edge".   Most customers only use px-dev for prototyping, since it has a limit of 3-nodes per cluster.    So using "edge" is perfectly acceptable.

Please move forward with this PR, or let us know what needs to be addressed.   Thank you.

jsilberm:
Ping?   GTG?

PR #428, opened 2017-02-10
Fix logmatic-docker entrypoint

joaquindlz:
1.0 (deprecated): NodeJS docker client, not compatible with the Logmatic.io integration.
1.2 (latest): use python.


rawmind0:
Hi @joaquindlz ... Thanks so much for the PR and sorry for the delay...

LGTM...

Issue #314, opened 2016-10-14
Add Minio to the community-catalog


blandes:
Nothing too crazy here. Just making a catalog template for [Minio](https://www.minio.io/) which is a minimalistic Amazon S3 compatible object storage server.

disaster37:
Hi, I will work on this enhance

zoidbergwill:
We're hoping to test out running this on Rancher soon. Any way I could help?

mathieu-gilloots:
:+1:

disaster37:
I work actively on it, i hope to do PR next week

PR #436, opened 2017-02-21
Add Minio Cloud Storage on Cattle community Catalog

disaster37:
Just add Minio Cloud Storage on community catalogue.
For more detail, look https://minio.io/

It's link with #314


templates/minio/0/rancher-compose.yml
```diff
@@ -0,0 +1,61 @@
+.catalog:
+  name: Minio
+  version: 2017-02-16-rancher1
+  description: |
+    Store photos, videos, VMs, containers, log files, or any blob of data as objects.
+  minimum_rancher_version: v0.59.0
+  maintainer: "Sebastien Langoureaux <linuxworkgroup@gmail.com>"
+  uuid: minio-0
```
cloudnautique: no longer needed.

cloudnautique:
Cool! I'm excited for this one.

Can you convert the docker-compose.yml to v2 syntax? Instead of the volume container, lets switch to a volume with the local driver. This will make the minimum_rancher_version 1.2.0

Where are the Docker files? I would be curious to see them. Could you put a link in the README or something?

disaster37:
It's seems not work with v2 syntax, can you help me ?

truongdo:
Nice work!
With MINIO_SCALE option, it seems like you are trying to setup a [minio distributed](https://docs.minio.io/docs/distributed-minio-quickstart-guide) system, but all minio nodes in this setup share the same docker volume. It does not make sense to run distributed system with a single volume.

Would you explain a bit more about this? Maybe I miss some points.

disaster37:
I think it's because to migrate from docker-compose v1 to docker-compose v2, it's change the pattern, and you are right it's make no sense.

The original docker-compose / rancher-compose create containers volumse for each minio container (sidekick) as volume-from.

truongdo:
Thanks for making it more clear. I still have some more comments on this.

Distributed minio are meant to be installed on different physical hard disks, not different docker volumes,
so even if you set it up with docker compose v1 with container volumes, it will not get any benefit from minio distributed system.

I also think it is not a good choice to use docker volumes, because docker volumes are (in most cases)
located in /var/lib/docker/volumes, which leads to 2 disadvantages:

1. It is inconvenient to mornitor and keep the data persistence. The data can be destroyed by removing the rancher stack and executing `docker volume rm $(docker volume ls -f dangling=true -q)` command.
It is easy to accidentally destroy the data without knowing so because the `docker volume rm` above does not tell us which volumes are being removed.

2. When users want to setup a distributed system, they really mean to keep the data safe and HA. Therefore, they will more likely have multi hosts and have the data partition separated from the boot partition. If you use docker volumes, all data will be stored in the boot partition unless you manually configure them not to.

Setting up an automate distributed minio system is a bit tricky, I think it would be better to,
1. Use host volumes instead, users can specify which disks they want to mount. This can prevent destroying the data accidentally, the only way to remove it is to run `rm folder_path` deliberately.
2. Setup a non-distributed system, this is ok for most basic use cases.  If users want to setup a distributed system, they need to do it manually so that they'll really understand how their data will be stored.

disaster37:
You are right, for my opinion, community catalogue it's just for test purpose. For production, you need to add scheduler constraint (one minio per host) and you need to add your persistent storage (with storage driver or volume path)

cloudnautique:
I need to play with this some, but I feel like we have a way to emulate the volume container with the top level volume declaration in V2. In general, community catalog entries should work in the most basic case, and offer configuration to deploy into more sophisticated setups. 

There are two dimensions that Minio scales as mentioned above. The single volume multi-node story is more compelling since it provides HA and redundancy. Then adding multiple disks and multi nodes is going to be the next most powerful. Though a single node with multiple disks should also be supported. 

In Rancher 1.5 we are introducing compose templating of the files which should open up some powerful flexibility. In the case of this template I can see it providing two key elements, volume driver type and number of volumes. It ought to give us the flexibility to also do host bind mounts as well.

truongdo:
> In Rancher 1.5 we are introducing compose templating of the files which should open up some powerful flexibility

What exactly improvement will be added?
If it is possible to define, add, or duplicate services in docker-compose.yml dynamically, it would make things much more flexible :). Let's say we have 4 hosts, then we can accomplish the distributed minio easily by duplicating the minio service to 4, then configure their scheduling label appropriately.

cloudnautique:
https://github.com/joshwget/rancher.github.io/blob/c35d7229f7a8129b1ca566fcb2695d72d6471a21/rancher/v1.5/en/cli/variable-interpolation/index.md

truongdo:
Oh, very nice!!
Is it possible to do templating at the service level? For example,
```
services:
{{- if eq .Values.WEB "true" }}
  web:
    image: nginx
    ports:
    - 8000
{{- end }}
```

disaster37:
Hi, 

have you any news about the volume data storage when service is scaled ?
Maybee I can doing a rollback to continue to use docker container data as sidekick ?

cloudnautique:
Hey, I took a shot at it. 

https://github.com/cloudnautique/community-catalog/tree/feature/minio

Its done off the work you have done. It deploys, but it doesn't seem to add the additional disks.

disaster37:
I look your problem,

You need to replace in environement `MINIO_DISKS_1=disk1` with:

```
{{- range $idx, $e := atoi .Values.MINIO_VOLUMES_PER_NODE | until }}
      - MINIO_DISKS_{{$idx}}=disk{{$idx}}{{- end }}
```

cloudnautique:
Cool, do you want to finish it up? I tried to do a pr to your repo, but it wouldn't let me. You did all the hard parts :)

disaster37:
Oki, I need to upgrade my Rancher before to test with your work base (new templating).


________________________________
De : Bill Maxwell <notifications@github.com>
Envoyé : vendredi 17 mars 2017 15:51
À : rancher/community-catalog
Cc : disaster37; Author
Objet : Re: [rancher/community-catalog] Add Minio Cloud Storage on Cattle community Catalog (#436)


Cool, do you want to finish it up? I tried to do a pr to your repo, but it wouldn't let me. You did all the hard parts :)

disaster37:
I need some help with templating and volumes.

It seems miss understanding this:
```
volumes:
  {{- if not (contains .Values.VOLUME_DRIVER "/")}}
    {{- range $idx, $e := atoi .Values.MINIO_DISKS | until }}
  minio-data-{{$idx}}:
    per_container: true
    driver: {{.Values.VOLUME_DRIVER}}
    {{- end}}
  {{- end}}
```

Because of, after create stack I get this on settings:
```yaml
volumes:
  minio-data-3:
    external: true
    driver: 'null'
  minio-data-2:
    external: true
    driver: 'null'
  minio-data-1:
    external: true
    driver: 'null'
  minio-data-0:
    per_container: true
```

I also look that scale: ${MINIO_SCALE} no continue to works on rancher-compose.yml

amy:
Hey @disaster37 what is your expected docker-compose.yml and rancher-compose.yml output? Would love to help you out with templating.

amy:
Hey @disaster37! For this part: `{{- if not (contains .Values.VOLUME_DRIVER "/")}}` I would recommend that you use `local` as a default.

Currently checking out the templating issues as well.

disaster37:
hey @amy, the logic is to use local as default, but the user can also use mount point on local node and so doesn't use storage volume.

amy:
@disaster37 Sorry for the late reply! In your rancher-compose.yml file, `minio-server:`should be nested under `services:`

That should populate `${MINIO_SCALE}`.

Looking at the docker-compose.yml now.

amy:
Ah. So in your docker-compose.yml [line 44](https://github.com/disaster37/community-catalog/blob/8749c4305b685e832cbf54cd2644cb7c0ddb8ccf/templates/minio/0/docker-compose.yml#L44), you have `{{.Values.VOLUME_DRIVER}}`. It should be `${VOLUME_DRIVER}`.

Let me know if you come across any issues with those changes.

disaster37:
Thx @amy, it' fix all issues. But now I try to add Loadblancer and publish minio service or loadblancer if user should this.
I doesn't found the right way to add Loadbalancer, I get the error on stack creation:

```
Error (Bad response statusCode [422]. Status [422 status code 422]. Body: [baseType=error, code=InvalidReference, fieldName=LaunchConfigName] from [http://localhost:8080/v2-beta/projects/1a152/services])
```

Any idea about that ?

amy:
hmmm... @disaster37 taking a peek now.

amy:
Another note. In terms of drone checks passing, whenever you have templating in your `.yml` files you need to append`.tpl`

disaster37:
Thx @amy, maybee it's better for drone test to append .tpl, but after that the calag entry not work.

disaster37:
Hum, so sorry, finnaly it's works as expected. Problem with my catalog refresh.
Can you confirm me that templating is not supported on rancher-compose.yml ?

Thx @amy, @cloudnautique and @truongdo. The catalog entry seems works as expected.

amy:
Yup! So interpolation exists for both the docker & rancher compose yml files. And templating exists just for docker-compose.yml

Glad to hear that you sorted out your load balancing issue before!

amy:
@disaster37 A suggestion for your volumes set up, your local volumes should be defined under the top level `volumes:` section. If possible, it would be best to avoid `volumes_from` and make it more explicit.

disaster37:
@amy you are right, thank for your suggestion ;)

mathieu-gilloots:
really needed feature :)

amy:
@disaster37 Awesome! Just tested it out and it works fabulously. Thanks for your work + cooperation!

templates/minio/0/rancher-compose.yml
```diff
@@ -0,0 +1,113 @@
+version: '2'
+catalog:
+  name: Minio
+  version: 2017-03-16-rancher1
+  questions:
```
cloudnautique: add: `minimum_rancher_version: v1.5.0` since this is using templates

cloudnautique:
@disaster37 we just need to add a minimum rancher version key so that only 1.5 users get this. Since it depends on templates.

disaster37:
I have put this on config.yml. I need to remove ot it ?

cloudnautique:
remove from the config.yml, it goes on the catalog section in rancher-compose.yml.

cloudnautique:
Thanks @disaster37 and @amy !

PR #440, opened 2017-02-25
fix(type): update field type string to password for secrets

ellerbrock:
gogs template


rawmind0:
Hi @ellerbrock thanks for the PR.....

LGTM...

PR #443, opened 2017-02-28
InterouteVDC driver added.

radu-stefanache:
Adding the Interoute VDC machine-driver.


machine-templates/interoutevdc/0/uiUrl
```diff
@@ -0,0 +1 @@
+https://raw.githubusercontent.com/Interoute/ui-driver-interoutevdc/v1.0/dist/component.js
```
vincent99: This doesn't work... github returns `Content-Type: text/plain` and browsers will refuse to interpret as JS.  Also only the domain that the driver is loaded from is automatically whitelisted, so the user would need to add `raw.githubusercontent.com` to the `api.proxy.whitelist` to allow it to be loaded.

vincent99:
@radu-stefanache 

image

radu-stefanache:
@vincent99  It works for me, I've tried it from multiple public facing servers (various zones/countries). Could be your browser/cors settings (if using)/vpn ?

vincent99:
That's a HTML response from the server (IIS), browser and CORS not relevant.  I didn't change anything but it's working now.

PR #451, opened 2017-03-09
Add rancher-haproxy template

mrproper:
Rancher-haproxy is: A haproxy server that uses the rancher-metadata service to do service discovery to route requests via hostname to stacks.

eg this will allow you to create a stack called 'foobar' and have requests hit http://foobar.$domain/ that will route to containers that have a specified label.
This is useful if you for example take an N stacks approach to CI/CD. ie. deploy app build 123 to a stack called 'app-r123'
This will give you an immediate endpoint http://app-r123.$domain'

The label to associate and the domain are configurable

References:
https://hub.docker.com/r/nodeintegration/rancher-haproxy/
https://github.com/nodeintegration/rancher-haproxy/


rawmind0:
Hi @mrproper ... Thanks so much for the PR...

The idea is so good, similar to traefik package scope, but you could do almost the same with rancher builtin haproxy and using selectors...
http://rancher.com/docs/rancher/v1.6/en/cattle/adding-load-balancers/#selector

Also, certificates would be integrated by this way, http://rancher.com/docs/rancher/v1.6/en/cattle/adding-load-balancers/#certificates

What do you think?? :)

cjellick:
v

PR #455, opened 2017-03-13
PostgreSQL fix: LB is set to HTTP protocol

janeczku:
Apparently this happens because of the legacy load balancer image.
Fixed by changing LB image to `rancher/lb-service-haproxy`.


rawmind0:
Hi @janeczku ... Thanks for the PR, good catch up.... :)

LGTM....

rawmind0:
Solves https://github.com/rancher/community-catalog/issues/514

PR #457, opened 2017-03-14
rancher supported version of ECR updater

wjimenez5271:


wjimenez5271:
Still need to confirm the docker image name used is going to be what @cloudnautique pushes to Dokcer Hub. @deniseschannon also planned to add some verbage to the README

PR #458, opened 2017-03-16
Add Artifactory OSS template

cthiebault:
Artifactory is a universal Binary Repository Manager.

https://www.jfrog.com/open-source/


templates/artifactory-oss/0/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+version: '2'
+
+services:
+
+  artifactory:
+    image: docker.bintray.io/jfrog/artifactory-oss:5.1.3
+    ports:
+     - "8081"
```
rawmind0: Need to set outside exposed port. If not, getting random port from the automatic range...
```
--     - "8081"
++     - "8081:8081"
```

Or better, set a rancher-compose HTTP_PORT variable to be configurable at deploy time...
```
--     - "8081"
++     - "${HTTP_PORT}:8081"
```

rawmind0:
Hi @cthiebault ... Thanks so much for the PR....

Could you please take a look at change request?? 

Best regards...

templates/artifactory-oss/0/docker-compose.yml
cthiebault: Thanks for the review :)
What if we don't want to map port but use a load balancer to serve Artifactory over HTTPS?

rawmind0: If you want to use lb instead expose port directly (recommended way), remove ports from artifactory, and add lb instance. Also, if you want the user could choose, you could add a variable in racnher-compose.yml and use go templating in docker-compose.yml.tpl 

As example, take a look at https://github.com/rancher/community-catalog/blob/master/templates/gocd/0/docker-compose.yml.tpl

cthiebault:
Sorry @rawmind0 I fucked up my fork :(
I've created a new PR with changes you requested: https://github.com/rancher/community-catalog/pull/549

PR #469, opened 2017-04-03
apache catalog

amy:


vincent99:
Can you add `.DS_Store` to `.gitignore` and remove the file from the PR?  (They are annoying binary files created by macOS to keep track of window position/style/etc)

amy:
@vincent99 whoops.

PR #473, opened 2017-04-04
Added new tag names

langhard:


rawmind0:
Hi @langhard ... thanks so much for the PR....

LGTM...

PR #474, opened 2017-04-04
Apache

amy:


cloudnautique:
Need a logo file called catalogImage-apache so that it shows up in the UI. Ideally it would be SVG followed by png.

cloudnautique:
The questions really ballooned, is there a way to simplify. I am not sure its worth all the questions to generate a self signed cert. At that point it is easier to just generate themselves.

Thoughts?

paulobezerr:
IMHO is good to change item name from Apache to "Apache PHP" for example.
You can use Apache to do a lot of different stuff and many times without PHP.

cjellick:
Closing this due to staleness. Feel free to open a new PR if there's still a relevant change to be made. Thanks!

PR #476, opened 2017-04-05
Add GoCD stack on version 17.3.0

disaster37:
I refactor the existing stack with many enhance (and more clear for my opinion):

- All is on the same stack with different service: one service for agent and one service for server.
- When you deploy the stack, you can choose to only deploy server or agent or the both.
- When you deploy the stack, you can choose to deploy Loadblancer and or expose the GoCD server.

**Note about GoCD server**:
- GoCD server is deployed with many usefull plugin if you choose to deploy extra plugins
- GoCD server is deploy with basic authentification
- You can choose the memory allowed to GoCD server
- You can choose the volume driver or the path mapped to the volume

**Note about GoCD agent**:
The goal is to have generic agent that support Docker. So you need to use dedicated docker image to build and test your stuff.
- GoCD Agent support Docker, Docker-compose, Rancher-compose and make
- GoCD Agent is deploy with Docker engine as sidekick
- You can choose the memory allowed to GoCD agent
- You can choose the number of GoCD agent you need to scale
- You can choose the server URL
- You can choose the volume driver or the path mapped to the volume


rawmind0:
Hi @disaster37 ... Very nice refactor, good job!! thanks so much.. :)

LGTM...

PR #479, opened 2017-04-07
fixing 3 images names

diegodorgam:
the images names were updated, this will fix the template for the services elasticsearch, postegresql and nginx.


rawmind0:
Hi @diegodorgam for the PR and contribute...

@monotek , thanks for the advice about duplicate. 
It's not exactly a duplicate of https://github.com/rancher/community-catalog/pull/449 . This modify the already created package version, that it's not working right now, but  #449 add a new version. 

LGTM

PR #488, opened 2017-04-19
Update GoCD template

disaster37:
Hi,

I found a bug and fix it when instanciate GoCD Template with path to to store persistance data (replace local by /mnt/gocd-server for exemple).
The `hasPrefix` function seems not working on templating. I replace it by ` if eq (printf "%.1s" .Values.VOLUME_DRIVER_SERVER) "/"` and all work as expected.


rawmind0:
Thanks so much @disaster37 ...

LGTM..

PR #489, opened 2017-04-19
Update Minio template

disaster37:
Hi,

I found a bug and fix it when instanciate Minio Template with path to store persistance data (replace local by /mnt/minior for exemple).
The hasPrefix function seems not working on templating. I replace it by if eq (printf "%.1s" .Values.VOLUME_DRIVER) "/" and all work as expected.


cloudnautique:
Can you resolve the merge conflicts? probably just need to rebase...

PR #493, opened 2017-04-26
Update Minio template

disaster37:
Hi,

I found a bug and fix it when instanciate Minio Template with path to store persistance data (replace local by /mnt/minior for exemple).
The hasPrefix function seems not working on templating. I replace it by if eq (printf "%.1s" .Values.VOLUME_DRIVER) "/" and all work as expected.

Replace #489


rawmind0:
Hi @disaster37 ... Thanks so much for the PR....

LGTM....

PR #499, opened 2017-05-03
Add logicmonitor collector catalog template

woz5999:
Add template for LogicMonitor collector container


templates/logicmonitor/0/docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+logicmonitor-collector:
+  image: logicmonitor/lm_collector:latest
```
rawmind0: Please, use concrete version instead latest....Latest is no reproducible

templates/logicmonitor/0/rancher-compose.yml
```diff
@@ -0,0 +1,36 @@
+.catalog:
+  name: LogicMonitor Collector
+  version: latest
```
rawmind0: Please, use concrete version instead latest....Latest is no reproducible

templates/logicmonitor/config.yml
```diff
@@ -0,0 +1,7 @@
+name: LogicMonitor
+description: SaaS-based performance monitoring for modern IT infrastructure
+version: latest
```
rawmind0: Please, use concrete version instead implicit latest....Latest is no reproducible

rawmind0 (changes requested):

rawmind0:
Hi @woz5999 ... thanks so much for the PR....

Could you please take a look at change request??

Thanks....

PR #500, opened 2017-05-03
Traefik-feat: go templating for add acme optional volume for persistence

rawmind0:


rawmind0:
Tested and working....

LGTM..

PR #508, opened 2017-05-16
Update Sematext catalog docker-compose.yml

tgjamin:
The docker-compose.yml would not get the proper `skip_by_image` or `skip_by_name` if specified when deploying. The rancher template names were accidentally specified from the "match" fields.


rawmind0:
Hi @tgjamin ... Thanks so much for the PR... :)

LGTM....

PR #513, opened 2017-05-25
add rancher max versions to k8s templates

deniseschannon:


deniseschannon:
https://github.com/rancher/rancher/issues/8837

PR #525, opened 2017-06-09
feat(templates): add openproject to rancher catalog templates

ellerbrock:
# OpenProject

Add OpenProject to the Rancher Catalog Templates.

woop woop


rawmind0:
Hi @ellerbrock ... Thanks so much for the PR....

LGTM...

PR #529, opened 2017-06-15
improve wekan catalog

Youkoulayley:
add questions to wekan installation


Youkoulayley:
tested on a local catalog : OK

PR #526, opened 2017-06-10
Update es-cluster to 5.4.2

daddz:

PR #534, opened 2017-06-22
Add ELK v5.4.1 stacks

dm:
Hi all,

I'd like to bring this full ELK stack (with latest v5.4.1) into the community catalog (actually two stacks; a simpler suitable for development, and a fully scalable version).

More details on my [own project repo](https://github.com/dm/rancher-catalog-elk).

I'm sure this might need some improvements, so feel free to give some feedback.

Cheers,
Dan

### DIFF
--- a/templates/elk-scalable-stack/0/docker-compose.yml.tpl
+++ b/templates/elk-scalable-stack/0/docker-compose.yml.tpl
@@ -1,0 +1,250 @@
+version: '2'
+services:
+  cerebro:
+    image: 'dmacedo/cerebro:0.6.5'
+    labels:
+      io.rancher.container.hostname_override: container_name
+    links:
+      - 'elasticsearch-data:elasticsearch-data'
+    stdin_open: true
+    tty: true
+
+  curator:
+    environment:
+      - 'ELASTICSEARCH_HOST=elasticsearch-data'
+      - 'ELASTICSEARCH_PORT=9200'
+      - 'MAX_INDEX_AGE=15'
+    image: sxmichael/curator-docker
+    links:
+      - 'elasticsearch-data:elasticsearch-data'
+    stdin_open: true
+    tty: true
+
+  elasticsearch-client:
+    cap_add:
+      - IPC_LOCK
+    depends_on:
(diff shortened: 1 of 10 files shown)


rawmind0:
Hi @dm ... Thanks so much for the PR....

Is not the already existing version of es v5.4.2 useful for you??  https://github.com/rancher/community-catalog/pull/526

Otherwise, you don't need to different versions of the package in order to add loadbalancer or not. Just use a variable and docker-compose.yml.tpl just to deploy or not loadbalancer. Same for non scalable version. Could you unify it in just one version with different variables/options?? 

Thanks....

dm:
Hi @rawmind0,

Note that this is different from an ElasticSearch-only cluster and if focused on the whole **ELK** stack (ElasticSearch, Logstash, Kibana), as well as a few extras for these: Logstash (Collector & Indexer), Logspout, Redis, Cerebro and Curator. So I wanted these to be plugged in.

The load balancer was really only need for testing and a specific deployment; and can be removed to bring this in line with remaining catalog items; I can care for this, as well as bring these up to the latest v5.4.3, as soon as I have a bit of time.

Cheers,
Dan

rawmind0:
@dm got it the target of the package... 

could you please, unify the 3 different versions into one with the name elk-stack and using docker-compose.yml.tpl ??  As an example, please take a look at https://github.com/rancher/community-catalog/blob/master/templates/gocd/0/docker-compose.yml.tpl 

Best regards...

cjellick:

dm:
> @dm got it the target of the package...
> 
> could you please, unify the 3 different versions into one with the name elk-stack and using docker-compose.yml.tpl ?? As an example, please take a look at https://github.com/rancher/community-catalog/blob/master/templates/gocd/0/docker-compose.yml.tpl

@rawmind0 @cjellick totally missed the notification on the above, and only changed stuff that we needed.
I haven't touched this in ages, so thanks for closing.

If I need to pick up ELK again for Rancher I'll be sure to reopen and update according to the above suggestion.

Cheers
Dan

PR #402, opened 2017-01-17
feat(add config variables for debug & loop for template janitor):

ellerbrock:
# Janitor Update

Add Configuration Variables for DEBUG and LOOP

lint rule: no pr without a gif :)

PR #547, opened 2017-07-10
Feat: Janitor add debug and loop options

rawmind0:
Based on https://github.com/rancher/community-catalog/pull/402 
Implemented change requests.


rawmind0:
Tested and working... LGTM...

PR #548, opened 2017-07-11
Upgraded traefik package to v1.3.3

rawmind0:


rawmind0:
Tested and working...

LGTM...

PR #557, opened 2017-07-21
Add mysql catalog entry

flaccid:
A simple catalog entry for a mysql stack using official mysql image as default.


templates/mysql/0/docker-compose.yml.tpl
```diff
@@ -0,0 +1,42 @@
+version: '2'
+services:
+  mysql-lb:
+    image: rancher/lb-service-haproxy
+    links:
+      - mysql
+    ports:
+      - ${mysql_lb_port}:3306
```
rawmind0: Please, use lb-service-haproxy version.
Link to the service is not needed anymore.
Port definition.
```
-    image: rancher/lb-service-haproxy
+    image: rancher/lb-service-haproxy:v0.6.4
-    links:
-      - mysql
    ports:
-      - ${mysql_lb_port}:3306
+      - ${mysql_lb_port}:${mysql_lb_port}
```

rawmind0:
Hi @flaccid .. thanks so much for your PR and contribute... :)

Please, take a look at changes request.

templates/mysql/0/docker-compose.yml.tpl
flaccid: - Added https://github.com/rancher/community-catalog/pull/557/commits/69fe7ae6102ac988da88f2481b93f9d117337bee for the haproxy image version.
- The link is kept for semantics e.g. when looking at the link graph in the UI we can see that the two services have a networking relationship
screen shot 2017-07-24 at 9 27 15 am
- `${mysql_lb_port}:${mysql_lb_port}` - I believe this is incorrect; the variable, `mysql_lb_port` is only the public port used by the lb; the backend port will always be 3306 within the container where mysqld listens; if the user changes it in this case, it won't work

rawmind0: - Thanks to add lb version.
- Link for semantics is ok. :)
- Lb doesn't work like that if user change the port. Here you need to define just the public port, not the backend one. The backend port is defined as  ```target_port: 3306``` in rancher-compose.yml . Please, change it  as
```
    ports:
-      - ${mysql_lb_port}:3306
+      - ${mysql_lb_port}:${mysql_lb_port}
```

rawmind0:
Thanks @flaccid , please, take a look at changes request.

flaccid:
@rawmind0 got it thanks. Can you jump on rancher users slack for some realtime? Wanna get this closed out due to timezone gap :)

rawmind0:
Hi @flaccid ... I'm already on rancher users slack... 
 Once you make changes requested, i'll test and merge it.  :)

templates/mysql/0/docker-compose.yml.tpl
flaccid: Pushed, thanks!

rawmind0:
Thanks @flaccid ....

LGTM...

PR #560, opened 2017-07-21
Added storage driver choice

joshuacox:
Seems to me that this should become the default and give people the choice of using an external NFS etc


joshuacox:
this has been contaminated with some other stuff, closing and resubmitting

PR #562, opened 2017-07-21
Added volume name and driver questions

joshuacox:
The variable VOLUME_DRIVER was not being asked, and I added a VOLUME_NAME as well


rawmind0:
Hi @joshuacox ... thanks for the PR....

LGTM...

PR #565, opened 2017-07-25
Reactioncommerce

joshuacox:
Adding the Reaction Commerce suite


rawmind0:
Hi @joshuacox ...thanks for the PR and contribute...

Difference between 2 versions is to use internal or external mongodb, isn't it?? 

Could you please, use go templating in docker-compose to differentiate them in one joined version??

something like that....docker-compose.yml.tpl
```
version: '2'
services:
  reaction:
    image: reactioncommerce/reaction:v1.4.0
    restart: always
    labels:
      io.rancher.scheduler.affinity:host_label: ${host_label}
      traefik.enable: true
      traefik.alias: ${REACTION_HOST}
      traefik.domain: ${REACTION_DOMAIN}
      traefik.acme: true
      traefik.port: 3000
    environment:
      MONGO_URL: "mongodb://mongo/reaction"
      ROOT_URL: "http://${REACTION_HOST}.${REACTION_DOMAIN}"
      REACTION_EMAIL: ${REACTION_EMAIL}
      REACTION_USER: ${REACTION_USER}
      REACTION_AUTH: ${REACTION_AUTH}
{{- if eq .Values.mongo_link ""}}
    external_links:
      - ${mongo_link}:mongo
{{- else}}
  mongo:
    command: mongod --storageEngine=wiredTiger
    restart: always
    environment:
      MONGO_SERVICE_NAME: mongo
      CATTLE_SCRIPT_DEBUG: ${debug}
    tty: true
    image: mongo:3.4
    labels:
      io.rancher.scheduler.affinity:host_label: ${host_label}
      io.rancher.container.hostname_override: container_name
    volumes:
      - mongodata:/data/db
volumes:
  mongodata:
    driver: ${VOLUME_DRIVER}
{{- end}}
```

templates/reactioncommerce/1/rancher-compose.yml
```diff
@@ -0,0 +1,70 @@
...
+        type: "password"
+      - variable: "mongo_link"
+        description: |
+          MongoDB external service link
+          cluster.
+        label: "External stack/service"
+        default: ""
+        required: true
+        type: "service"
+      - variable: "MONGO_URL"
```
rawmind0: Sorry, but why this question if you advise to not change?? Setting the external link in the compose, MONGO_URL could be fixed as in the other version.

templates/reactioncommerce/1/rancher-compose.yml
```diff
@@ -0,0 +1,70 @@
...
+        required: true
+        default: ""
+        type: "password"
+      - variable: "mongo_link"
+        description: |
+          MongoDB external service link
+          cluster.
+        label: "External stack/service"
+        default: ""
+        required: true
```
rawmind0: Not required if you use go templating in docker-compose.

templates/reactioncommerce/1/rancher-compose.yml
joshuacox: well if you don't set the external link and change the MONGO_URL you can get this stack to utilize a mongoDB outside of rancher altogether, which I am doing at the moment on one of my tests

joshuacox:
Yes the two different versions are for external link versus sticking as closely to the reaction official docker-compose.yml.  I personally use the second version with external links to the exclusion of the official method, but I feel that it deserves historical context.

My question then is if you use go templating to in docker-compose.yml can I do the same with the services section in rancher-compose.yml?

```
services:
{{- if eq .Values.mongo_link ""}}
  mongo:
    scale: 1
    retain_ip: true
{{- end}}
```

rawmind0:
Good question @joshuacox .. No, in rancher-compose you couldn't use go templating, but in that case you don't need... 

Services defined in rancher-compose that doesn't exist in docker-compose are ignored... :)
```
services:
  mongo:
    scale: 1
    retain_ip: true
```

templates/reactioncommerce/1/rancher-compose.yml
rawmind0: But in that case, mongo_link is required. you never get the case of don't set external link.

Anyway, if you want to use an external database, may be it would be better to use external service, and use same mongo_url...
```
  mongo:
    external_ips:
    - 1.1.1.1
    start_on_create: true
```

joshuacox:
I think you got that backwards? In that services defined in docker-compose and not in rancher-compose are ignored.  I am pretty sure that defined in rancher-compose and not in docker-compose results in an image undefined error.

joshuacox:
I'm kind of on the fence about the templating, without it in the rancher-compose as well it means I have to ask questions that are irrelevant some of the time depending on how you answered other questions, perhaps I should just simplify it all to clear confusion on this 'demo' catalog and let people make more templates to cover their custom situations in their own custom catalogs.  Like I do [here](https://github.com/ohmydocker/ohmydocker-catalog/tree/master/templates/reactioncommerce-extdb)

joshuacox:
what do you think about removing 0 and 1 and leaving the third option as the only way to go about it, with the main points being

uses external linked mongo stack in rancher
allows the user to rename the db so that multiple instances can use the same mongo stack

rawmind0:
I'm just testing the docker-compose.yml.tpl version that i posted, and both options are working, "internal" and "external" mongo ...

Please, test the version 1.4.0-rancher2 from here, https://github.com/rawmind0/service-catalog

joshuacox:
Yes your templated version is a very elegant way of combining the two.  I added in the DB name choice as well so people can have multiple instances using the same external link mongo stack.  And for now we can forget about external to rancher mongo

joshuacox:
Am I crazy for thinking services defined in rancher-compose.yml and not in docker-compose.yml?

It seems like I just ran into with another template I was working on?

templates/reactioncommerce/0/docker-compose.yml.tpl
```diff
@@ -0,0 +1,41 @@
...
+      traefik.acme: true
+      traefik.port: 3000
+    environment:
+      MONGO_URL: "mongodb://mongo/${MONGO_DB}"
+      ROOT_URL: "http://${REACTION_HOST}.${REACTION_DOMAIN}"
+      REACTION_EMAIL: ${REACTION_EMAIL}
+      REACTION_USER: ${REACTION_USER}
+      REACTION_AUTH: ${REACTION_AUTH}
+{{- if ne .Values.mongo_link ""}}
+      REACTION_TEST: ${REACTION_AUTH}
```
rawmind0: Sorry, my fault, this line was just for testing...
```
-      REACTION_TEST: ${REACTION_AUTH}
```

joshuacox: removed

rawmind0:
No, you are not crazy....To be honest, i was doubting about that, but i tested and works... :)

rawmind0:
Thanks for the commits.... :)

LGTM...

PR #571, opened 2017-07-31
change default to ignore *-datavolume containers in janitor:docker-cl…

wtayyeb:
change default to ignore *-datavolume containers in janitor:docker-cleanup


rawmind0:
Hi @wtayyeb ... thnas so much for the PR....

LGTM...

PR #569, opened 2017-07-27
0.2.1 rabbitmq

joshuacox:
Added 0.2.1 to accomodate upstream changes https://github.com/rancher/catalog-dockerfiles/pull/103 that PR is probably obsolete with this

PR #575, opened 2017-08-01
Fix rabbitmq-3 network definition and add optional host_label

rawmind0:


rawmind0:
Fix https://github.com/rancher/community-catalog/pull/569

Issue #580, opened 2017-08-04
Prometheus v5 data volume mismatch (no data persistence)


zanchin:
v5 of the catalog has the following config:
```
command:  ...  -storage.local.path=/prometheus ...
volumes:
 - /data/
```

So `/data` is being mounted as a volume, but all data is being stored in `/prometheus`, which means that Prometheus as it ships does not persist it's data!

rawmind0:
Thanks @zanchin .. 

Merged PR, closing issue.

PR #581, opened 2017-08-04
Persist prometheus and grafana volumes by default

zanchin:
see #580 for initial issue


rawmind0:
Hi @zanchin ... Thanks so much for your PR and contribute.... :)

LGTM....

Issue #395, opened 2017-01-06
QuasarDB category typo ?


radu-stefanache:
Shouldn't QuasarDB be in the databases category ?
Instead is in the Database (!=Databases) category.

https://github.com/rancher/community-catalog/blob/master/templates/quasardb-community/config.yml#L5

rawmind0:
thanks @radu-stefanache ... 

Fixed category to databases... https://github.com/rancher/community-catalog/pull/587

PR #587, opened 2017-08-11
FIX category to databases to influxdb, mysql, postgres and quasardb packages

rawmind0:


rawmind0:
Solve https://github.com/rancher/community-catalog/issues/395

PR #589, opened 2017-08-14
update dnsupdate to use custom dns record

wtayyeb:
By default DNS entries are named based on default template defined on dnsupdate service.
You can customize it per service by assigning `io.rancher.service.external_dns_name_template` label to each service.
dnsupdate will use this template instead of default one when creating dns record for related service.


rawmind0:
Hi @wtayyeb ... thanks so much for the PR.... :)

LGTM...

PR #592, opened 2017-08-16
Adding catalog icon + updated catalog attributes

wjimenez5271:


templates/iron-functions/config.yml
```diff
@@ -1,5 +1,9 @@
 name: IronFunctions
 description: |
-  Serverless platform using Docker
+  Serverless platform using Docker (https://github.com/iron-io/functions)
 version: v0.0.1
```
rawmind0: Could you please set correct default version?? 
```
-  version: v0.0.1
+  version: v0.0.2
```

rawmind0:
Hi @wjimenez5271 , thanks for the PR... :)

drone checks are failing due to package default version. Please take a look at change request.

templates/iron-functions/config.yml
wjimenez5271: ah, good catch, sorry still learning the structure of these

rawmind0:
Thanks @wjimenez5271 ...

LGTM

PR #599, opened 2017-08-22
Internal DNS fix

AleksandrNull:
Because container have host networking it wouldn't be able to access rancher internal DNS server so access to the ETCD by DNS/FQDN name wouldn't work.


rawmind0:
Hi @AleksandrNull ... thanks so much for the PR...

LGTM...

PR #605, opened 2017-08-27
add gitlab latest image version

mittz:
update GitLab catalog, the latest GitLab docker image version is available.


mittz:
there is another bug in this catalog, so i close this PR and retry after fixing the bug.

PR #611, opened 2017-09-01
drone typo fix for drone_admin

aisensiy:
Change the `drone_admins` variable to `drone_admin`. Without this change, the drone admin will be empty even if you typed one during creating the stack.


rawmind0:
Hi @aisensiy ....thanks so much for the PR... :)

LGTM...

PR #626, opened 2017-09-23
Wekan Update

joshuacox:
Updated the template to use version 2 and templatized mongo on whether an external link is provided, and added traefik labels so traefik will pick it up.


rawmind0:
Hi @joshuacox , thanks for the PR....

LGTM...

PR #635, opened 2017-10-04
Newlines inserted by es-cluster sysctl param breaks yml validation

superseb:
Tested the current implementation on v1.5.10 and v1.6.10 but both error out when launched from the UI.


rawmind0:
Hi @superseb ... Thanks for the PR.... :)

LGTM...

PR #642, opened 2017-10-18
upgrde rancher dns to 0.7.6 on gandi stack due to a breaking change on gandi's api

jeanlouisboudart:
see https://github.com/rancher/rancher/issues/10136


jeanlouisboudart:
@deniseschannon ping

PR #646, opened 2017-10-24
Update Jenkins template to jenkins/jenkins and support multi version

heralight:
Update Jenkins template to jenkins/jenkins and support version with question.

* lts
* lts-alpine
* ... etc.


rPawel:
I'd really appreciate if this PR would get merged. The current template is heavily outdated, and forced upgrades outside of the main pipeline are rather tedious.
Also, current _/jenkins has been deprecated and is not receiving any updates. This PR is switching to jenkins/jenkins which is a huge win.

heralight:
You can, just rename it, my  push was done before version 4 was created...

chrisurwin:
@rPawel the latest template in Rancher already uses jenkins/jenkins as the base.
@heralight We need to have specific versions in a template in order to allow for the catalog to check if there is an upgrade available.

heralight:
> @heralight We need to have specific versions in a template in order to allow for the catalog to check if there is an upgrade available.

I 'm not  talking about template rancher version, which is good and understand, I talk about template code like that:
` image: "jenkins/jenkins:2.92"`

which from my point of is useless because:
* fix version version
* cannot use or test a fork
* need maintenance

I prefer something like I push:
`image: "${jenkins_image}"` with a default value.
For me a template need to generic and open as more as we can... to fit any simple situation.

Actually, your jenkins template implementation need maintenance and will be outdated soon... then useless...

chrisurwin:
People don't necessarily want to check what version they should be running, so if you put a default how do you then notify of a change, if you update the default you are in the same position you are now.

The catalog is a simple way to deploy applications, more advanced users will tend to create their own private catalog and maintain it themselves, or take the rancher/docker-compose files and tweak it to their needs.

You can test a new version by creating a private catalog, which you can fork and the catalog service supports

PR #659, opened 2017-11-05
Update to v0.7.15 of Rancher's haproxy lb

DevKyleS:


DevKyleS:
I just did a simple find/replace, now realizing that perhaps it may be best to add a newer version to each application instead of updating existing scripts...
What do you think?

chrisurwin:
Each application would need to be split into a separate PR, and an update like this should create a new version for each application too.

PR #671, opened 2017-11-13
Update rancher ui

alibazlamit:
Update rancher component and docker driver URLs and versions

### DIFF
--- a/machine-templates/profitbricks/1/rancher-compose.yml
+++ b/machine-templates/profitbricks/1/rancher-compose.yml
@@ -1,0 +1,3 @@
+.catalog:
+  name: "profitbricks"
+  version: "v1.3.3"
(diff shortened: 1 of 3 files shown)


DevKyleS:
I would suggest moving these changes to a new folder "1" instead of modifying the contents in the 0 folder.

alibazlamit:
Updated with folder 1 for the newer version.

alibazlamit:
@DevKyleS  Sorry to ping you on this closed PR. I want to do the same PR but for Packet baremetal provider. We have the ui driver setup here https://github.com/packethost/ui-driver-packet and we would like to add the latest values to rancher 2 by default. Could you please point me to the location of rancher 2 machine templates.

Thanks

PR #684, opened 2017-11-29
Netdata template cosmetic fixes

genuss:
- If you don't mount docker socket from host netdata will not be able to understand human-readable names for cgroups, so all you containers will be ``aa123bc``.
- Also use host UTS namespace to set hostname of a container to an agent hostname. In the case of one container per host it will be easy to indentify netdata instances


chrisurwin:
This needs to be created as a new folder (/1) or existing users won't get the update template to see the names

genuss:
Done!

PR #699, opened 2017-12-09
Update traefik package to v1.4.5.

rawmind0:
- Updated traefik to v1.4.5
- Rancher external integration using rancher-template to get config update immediately.
- Fix ssl cert and key update.


rawmind0:
Tested and working...

LGTM

PR #702, opened 2017-12-15
Add Rancher Pipeline into Catalog

biblesyme:
For more usage and guide information about this Catalog. Please refer to https://github.com/rancher/pipeline.

Related Issue is [here](https://github.com/rancher/rancher/issues/10591)


infra-templates/pipeline/config.yml
```diff
@@ -0,0 +1,4 @@
+name: Rancher Pipeline 
+description: (Experimental) Rancher CI service
+version: 1.0.0
+category: Continuous Integration
```
deniseschannon: Will this work with any orchestration or is it `cattle` specific? Let's add the following labels assuming it only works in cattle. 

```
labels:
  io.rancher.certified: Experimental
  io.rancher.orchestration.supported: 'cattle'
```

gitlawr: updated

infra-templates/pipeline/0/rancher-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+      label: "Host with Label to put pipeline components on"
+      description: |
+        Run pipeline components on Host with specific labels. 
+        Example: 'pipeline=true'
+      required: false
+      default: ""
+      type: "string"
+services:
+  jenkins-slave:
+#{{- if ge .Values.SLAVES "1" }}
```
deniseschannon: why are there `#` in these lines?

infra-templates/pipeline/config.yml
```diff
@@ -0,0 +1,7 @@
+name: Rancher Pipeline 
+description: (Experimental) Rancher CI service
```
deniseschannon: No need to put in Experimental in the name a we have the experimental in the banner.

infra-templates/pipeline/0/rancher-compose.yml
gitlawr: It will fail due to syntax validation without `#`. This is the way I find to work out.

infra-templates/pipeline/config.yml
gitlawr: I'll remove that.

infra-templates/pipeline/0/rancher-compose.yml
deniseschannon: can you try and change it to `rancher-compose.yml.tpl`? does the templating actually get evaluated?

PR #714, opened 2018-01-06
Added external flag for environment scoped volumes

hprotzek:
To use an environment scoped volume the option `external: true` is required in the volumes section. If the option is missing, the volume is always stack scoped. The additional flag `ENVIRONMENT_SCOPED` fix the problem.

### DIFF
--- a/templates/letsencrypt/4/docker-compose.yml.tpl
+++ b/templates/letsencrypt/4/docker-compose.yml.tpl
@@ -59,7 +59,7 @@ volumes:
       {{.Values.STORAGE_DRIVER_OPT}}
     {{- end }}
     {{- end }}
-    {{- if .Values.ENVIRONMENT_SCOPED}}
+    {{- if eq .Values.ENVIRONMENT_SCOPED "true" }}
     external: true
     {{- end }}
 {{- end }}
--- a/templates/letsencrypt/4/docker-compose.yml.tpl
+++ b/templates/letsencrypt/4/docker-compose.yml.tpl
@@ -59,4 +59,7 @@ volumes:
       {{.Values.STORAGE_DRIVER_OPT}}
     {{- end }}
     {{- end }}
+    {{- if .Values.ENVIRONMENT_SCOPED}}
+    external: true
+    {{- end }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


templates/letsencrypt/4/docker-compose.yml.tpl
```diff
@@ -59,4 +59,7 @@ volumes:
       {{.Values.STORAGE_DRIVER_OPT}}
     {{- end }}
     {{- end }}
+    {{- if .Values.ENVIRONMENT_SCOPED}}
+    external: true
+    {{- end }}
```
rawmind0: Your 'if' statement is not correct. Due to variable is a boolean it always enter into the if. Have you tested it??

Also, like it's specified at https://docs.docker.com/compose/compose-file/compose-file-v2/#external, `external cannot be used in conjunction with other volume configuration keys (driver, driver_opts).`

```
-    {{- end }}
+    {{- else }}
-    {{- if .Values.ENVIRONMENT_SCOPED}}
+    {{- if eq .Values.ENVIRONMENT_SCOPED "true" }}
    external: true
    {{- end }}
+    {{- end }}
```

rawmind0 (changes requested):

rawmind0:
Hi @hprotzek, thanks for the PR but it's not working.

Please, take a look to requested changes.

hprotzek:
Updated the boolean statement. Not sure how to comply with the docker compose requirement, because the Rancher doc say's different: [http://rancher.com/docs/rancher/v1.6/en/rancher-services/storage-service/#environment-scoped-1]
If I only add external, how does Rancher knows about the required driver?

rawmind0:
If you set `external: true`, it's supposed that volume was already created and exists at docker daemon. Docker just tries to look for it by name, driver and driver_opts are implicit, https://docs.docker.com/compose/compose-file/compose-file-v2/#external
`If set to true, specifies that this volume has been created outside of Compose. docker-compose up will not attempt to create it, and will raise an error if it doesn’t exist.`

Rancher doc you are referencing should be wrong. Thanks, i'll change it.

rawmind0:
https://github.com/rancher/rancher.github.io/pull/969

rawmind0:
Please, implement `else` part of the changes requested.

cjellick:

PR #717, opened 2018-01-09
Feat: Updated traefik external rancher integration

rawmind0:


rawmind0:
Tested and working.

LGTM...

PR #720, opened 2018-01-16
Update minio to 2018-01-02_1

mschneider82:
Security fix release

Also supporting TLS_FQDN for optional TLS. By settings this to a Value a selfsinged certificate will be created. This is required to support service side encryption.


rawmind0:
Hi @mschneider82 , thanks for the PR...

LGTM...

PR #722, opened 2018-01-17
Aqua update

jerbia:
Changed the Aqua version to latest - 2.6.4.


rawmind0:
Hi @jerbia  , thanks for the PR...

In order, rancher users that use aqua package get notified about aqua upgrade, could you please add a new catalog version with the updated compose files??

Add new compose files to new `templates/aqua/1` folder defining a new package version.
Setting new version to templates/aqua/config.yml version field as you do.
Doing in that way, all rancher users that use aqua, get notified that there is an upgrade available. :)

rawmind0:
LGTM...

PR #729, opened 2018-02-02
hetzner: fixed dead link

mxschmitt:
Hi,
unfortunately I renamed my account yesterday, so the ui-driver has dead links. I'm not sure what better is, just a fix to the existing version or create a new one with the new url.
Best regards and sorry
Max


rawmind0:
Hi @mxschmitt , in that case better to fix existing version. Thanks for the update.

LGTM...

PR #730, opened 2018-02-02
Updated traefik package to v1.5.1

rawmind0:


rawmind0:
Tested and working.

LGTM...

PR #733, opened 2018-02-06
add filebeat for rancher catalog for fast deployment

easonlau02:
Hi guys,

Here to add filebeat 6.1.1 for rancher catalog for fast deployment.
Hope we can use at the rancher community catalog.
Please kindly help review. Thanks.

Regards/Eason

### DIFF
--- a/templates/filebeat/0/docker-compose.yml
+++ b/templates/filebeat/0/docker-compose.yml
@@ -1,0 +1,22 @@
+version: '2'
+services:
+  filebeat:
+    image: eason02/filebeat-alpine:5.3.1
+    container_name: filebeat-alpine-5.3.1
+    restart: always
+    labels:
+      io.rancher.container.pull_image: always
+    environment:
+      - env=${ENV}
+      - logstash=${LOGSTASH}
+      - project=${PROJECT}
+    network_mode: host
+    volumes:
+      - /var/lib/filebeat/:/filebeat/data
+      - ${APPLOG_PATH}:/home/user/
+      - ${NGINXLOG_PATH}:/var/log/nginx/
+    logging:
+      driver: json-file
+      options:
+        max-file: '5'
+        max-size: 10m
--- a/templates/filebeat/0/docker-compose.yml
+++ b/templates/filebeat/0/docker-compose.yml
@@ -11,8 +11,8 @@ services:
     network_mode: host
     volumes:
       - /var/lib/filebeat/:/filebeat/data
-      - ${APPLOG_PATH}:/home/user/
-      - ${NGINXLOG_PATH}:/var/log/nginx/
+      - ${APPLOG_PATH}:/home/user/:ro
+      - ${NGINXLOG_PATH}:/var/log/nginx/:ro
     logging:
       driver: json-file
       options:
--- a/templates/filebeat/0/docker-compose.yml
+++ b/templates/filebeat/0/docker-compose.yml
@@ -4,8 +4,6 @@ services:
     image: eason02/filebeat-alpine:5.3.1
     container_name: filebeat-alpine-5.3.1
     restart: always
-    labels:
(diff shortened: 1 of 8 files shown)


templates/filebeat/2/rancher-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      label: "Environment Name"
+      description: "Environment to distinguish data source"
+      default: "QA"
+      required: true
+      type: "enum"
+      options:
+        - QA
+        - PP
+        - INT
+        - PROD
```
rawmind0: It's nice to have predefined values, but what happens if my environments notation is different?? Mey would be better to use an string type to allow free input. What do you think??

templates/filebeat/1/rancher-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      label: "Environment Name"
+      description: "Environment to distinguish data source"
+      default: "QA"
+      required: true
+      type: "enum"
+      options:
+        - QA
+        - PP
+        - INT
+        - PROD
```
rawmind0: It's nice to have predefined values, but what happens if my environments notation is different?? Mey would be better to use an string type to allow free input. What do you think??

templates/filebeat/0/rancher-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      label: "Environment Name"
+      description: "Environment to distinguish data source"
+      default: "QA"
+      required: true
+      type: "enum"
+      options:
+        - QA
+        - PP
+        - INT
+        - PROD
```
rawmind0: It's nice to have predefined values, but what happens if my environments notation is different?? Mey would be better to use an string type to allow free input. What do you think??

templates/filebeat/0/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+version: '2'
+services:
+  filebeat:
+    image: eason02/filebeat-alpine:5.3.1
+    container_name: filebeat-alpine-5.3.1
+    restart: always
+    labels:
+      io.rancher.container.pull_image: always
```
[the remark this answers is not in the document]
easonlau02: Oh yes. which is my rancher docker-compose version. Remove at next commit.

templates/filebeat/0/rancher-compose.yml
```diff
@@ -0,0 +1,43 @@
+version: '2'
+catalog:
+  name: "filebeat-alpine"
+  version: "5.3.1"
+  description: "Catalog's rancher-compose"
+  minimun_rancher_version: v1.0.1
+  maximun_rancher_version: v1.6.12
```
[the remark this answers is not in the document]
easonlau02: Actually it supports. Updated later.

easonlau02:
Updated and passed checking. @rawmind0

rawmind0:
Hi @easonlau02 , thanks for PR and the changes. 

You are assuming that the package would be run at host level, forcing network_mode and bind volumes to get config and logs files. What do you think to add distinct volumes driver as bind, nfs,... to make the package more versatile?? May be better if ${APPLOG_PATH} and ${NGINXLOG_PATH} volumes would be mounted in read-only mode?? 

Also network-mode could be managed too if logstash would be running inside rancher. What do you think add hosts or managed option??

easonlau02:
hi @rawmind0, thanks for your fast review.

For ${APPLOG_PATH} and ${NGINXLOG_PATH} indeed need read-only. let me update.
But for network-mode, I don't think let it as managed, which can prevent from running with logstash as one host, so that someone maybe set it as localhost. Now let network-mode as host, which will support logstash host as localhost or server name or domain. What do you think about this??

rawmind0:
I get your point @easonlau02, thanks for the changes. 

LGTM...

PR #735, opened 2018-02-09
Fix healthcheck and add DB host configuration

shakefu:
- The current healthcheck is `/auth/login/sentry` which spams the logs with 301 redirects because Sentry expects a canonical trailing slash.
- Updates the healthcheck to `/_health/` which is [described in the docs](https://docs.sentry.io/server/monitoring/) as the correct way to verify the web app health.
- Adds a `sentry_db_host` configuration setting (which defaults to the existing `sentry-postgres` service), to allow for using an external database from initial setup.

### DIFF
--- a/templates/sentry/4/rancher-compose.yml
+++ b/templates/sentry/4/rancher-compose.yml
@@ -1,7 +1,7 @@
 version: '2'
 catalog:
   name: Sentry
-  version: 8.21.0
+  version: 8.21.0-1
   description: Sentry is a realtime event logging and aggregation platform
 
   questions:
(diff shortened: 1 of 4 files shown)


templates/sentry/4/rancher-compose.yml
```diff
@@ -0,0 +1,137 @@
+version: '2'
+catalog:
+  name: Sentry
+  version: 8.21.0
```
rawmind0: Package version field should be different from previous releases. 
Could you please change it??
```
-  version: 8.21.0
+  version: 8.21.0-1
```
Also, `templates/sentry/config.yml` version field should be updated.

rawmind0:
Hi @shakefu , thanks for the PR.

Could you please, take a look ate changes requested??

rawmind0:
Thanks for the changes @shakefu...

LGTM..

PR #741, opened 2018-02-19
Updated traefik to v1.5.2

rawmind0:


rawmind0:
Tested and working.

LGTM...

PR #748, opened 2018-03-02
Migrate to Prometheus 2.1.0 and Grafana 5.0.0

sliard:
Change Prometheus and Grafana version.
Will work only when infinityworks upgrade infinityworks/prom-conf with this pull request : 
https://github.com/infinityworks/prom-conf/pull/6

### DIFF
--- a/templates/Prometheus/6/README.md
+++ b/templates/Prometheus/6/README.md
@@ -1,0 +1,26 @@
+# Prometheus
+
+### Info:
+
+This template deploys a collection of monitoring services based upon the technologies listed below, once deployed you should have a monitoring platform capable of querying a wide variety of metrics that represent your environment, also included are somehandy pre-configured dashboards to get you started.
+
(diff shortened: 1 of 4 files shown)


Rucknar:
@sliard Looks good, it's probably worth updating the default version to match your new entry. Can be done in the main config file, you okay to push that up as an update?

sliard:
No problem. I just didn't want to force this update and let you choose. :)
For the moment I set it has version 3.0.1 but don't you think 4.0.0 is better ?
Let me know what you prefer and I will update.

Rucknar:
Won't force it on anyone. That version number needs adjusting in the base folder config.yaml file to update the default version. If we want to hold off on that for a little while though, that's fine.

Rucknar:
@cloudnautique Looks good to merge to me, haven't had chance to test myself.

rawmind0:
Hi guys, @sliard thanks for the PR, @Rucknar thanks for the review.

Tested and working fine. 

LGTM

PR #751, opened 2018-03-11
letsencrypt: Allow optional `external: true` in volume

bcardiff:
Allows the volume to have a `external: true` if needed so the certificate can be shared across stacks.
My use case for this is that the certificate is needed by a smtp forwarder container.

In case this is something wanted, since the change is backward compatible I am not sure whether is appropriate to overwrite the latest version or create a new version template.


bcardiff:
Hi @janeczku, I wanted to check if a notification got lost or if there something to discuss about this PR.

cjellick:
Closing this due to staleness. Feel free to reopen or open a new PR if there's still a relevant change to be made. Thanks!

PR #759, opened 2018-03-26
Updated traefik package to v1.5.4

rawmind0:
- Updated traefik to v1.5.4
- Added traefik timeouts to catalog question. https://docs.traefik.io/configuration/commons/#timeouts


rawmind0:
Tested and working...

LGTM...

PR #761, opened 2018-03-27
Allowing TTL to be specified by Env var

benyanke:
This should really be a configurable option

### DIFF
--- a/infra-templates/digitalocean-dns/3/README.md
+++ b/infra-templates/digitalocean-dns/3/README.md
@@ -11,7 +11,7 @@ Rancher External DNS service powered by DigitalOcean
 #### Usage
 
 ##### DigitalOcean DNS record TTL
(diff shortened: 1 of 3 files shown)


rawmind0:
Hi @benyanke , thanks for the PR and contribute.

Couple of questions, why are your updating files at `infra-templates/digitalocean-dns/0` and `infra-templates/digitalocean-dns/3` versions?? Why don't you update all versions or just the last one??

benyanke:
I previously misunderstood the versioning - I see the issue with that now. Is there an easy way to delete a commit from a PR or should I start over?

rawmind0:
You could do it with `git reset` in your forked repo. If not you could start over.

benyanke:
ok, force pushed my cleaned up branch - give it a look. Thanks @rawmind0

rawmind0:
Thanks @benyanke ..

LGTM...

PR #762, opened 2018-03-29
Feat: infoblox package option to use rancher secrets 

rawmind0:
Feat: infoblox package option to use rancher secrets instead env variable for password


rawmind0:
Tested and working...

LGTM

PR #764, opened 2018-04-05
Add MailHog v1.0.0

RonWilliams:
Adds support for https://hub.docker.com/r/mailhog/mailhog/.


templates/mailhog/0/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '2'
+services:
+  mailhog:
+    image: mailhog/mailhog:v1.0.0
+    environment:
+      MH_UI_WEB_PATH: ${mh_ui_web_path}
+      MH_SMTP_BIND_ADDR: ${mh_smtp_bind_addr}
+    stdin_open: true
+    tty: true
+    user: root
```
leodotcloud: Is "root" user really necessary?

templates/mailhog/0/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '2'
+services:
+  mailhog:
+    image: mailhog/mailhog:v1.0.0
+    environment:
+      MH_UI_WEB_PATH: ${mh_ui_web_path}
+      MH_SMTP_BIND_ADDR: ${mh_smtp_bind_addr}
```
leodotcloud: without any publish ports, how is this service accessible?

templates/mailhog/0/docker-compose.yml
RonWilliams: When running the container on SMTP's default port 25, the container must run as root due to permissions issues. Without the root user specified, the container will not run as expected. If you would like, I can open a "to-do" against the MailHog (https://github.com/mailhog/MailHog) project to see if they will accept a change permitting running on port 25 using the standard user. 

The use of root seems to apply in the case of https://github.com/rancher/community-catalog/blob/666a687a3a03b9a6923b796abb192c9ce04dd59d/templates/aqua/1/docker-compose.yml. Hoping we can get this approved and enhance in the future.

templates/mailhog/0/docker-compose.yml
RonWilliams: The service is used internally as a mail proxy SMTP port 25. Containers use an external_link to reference the container (sample below). The Web UI is exposed via a Rancher Load Balancer (LB); we do not expose the service directly in our use case as we do hostname based routing in Rancher LB.

It should be pretty easy to make an option to "expose publicly" or something like that. I avoided the implementation since we often experience significant issues using many of the existing catalog services due to their propensity to expose on common ports (ex. 80).

leodotcloud:
LGTM. @cloudnautique can you please review and merge?

templates/mailhog/0/docker-compose.yml
rawmind0: @RonWilliams, could you please add a simple `templates/mailhog/0/README.md` explaining it for user that would like to publish smtp and/or UI service (8025 tcp port) by rancher LB??

RonWilliams: @rawmind0 Good idea, added.

rawmind0:
Thanks @RonWilliams ...

LGTM

PR #767, opened 2018-04-08
add es 6.2.3 with x-pack and Microsft Sqlserver 2017

jagjag:
ADD  variable  elastic_password
Change  image  to docker.elastic.co/elasticsearch/elasticsearch-platinum:6.2.3
Change  ENV  "xpack.security.enabled=true"


rawmind0:
Hi @jagjag, thanks for the PR and contribute. 

Looking at PR, i think it's not the best way to add `x-pack` option for elasticsearch. 

Comments:
- Generate a package version with x-pack included. Could be better to add a catalog enum question, asking for elasticsearch version `normal` or `x-pack`, and set docker images for es, based on this question. Like it's done at artifactory-oss package with `ARTIFACTORY_VERSION`, https://github.com/rancher/community-catalog/blob/master/templates/artifactory-oss/0/rancher-compose.yml https://github.com/rancher/community-catalog/blob/master/templates/artifactory-oss/0/docker-compose.yml.tpl

- Couple kibana inside the package. Could be better just add same option at kibana package and not to couple here. 

- Add mssqlserver. That's cool package to add, but should be better in a standalone PR.

Could you please separate it into 3 distinct PR's?? One for elasticsearch, one for kibana, adding `x-pack` options, and one for mssqlserver.

jagjag:
ok I see

PR #769, opened 2018-04-11
Remove traefikdomain because it was never defined

### LINKED ISSUE #707: Can't console-login to last version 1.1.1-1 of the Harbor Registry.
docker login 10.8.0.1:80
Username: admin
Password: 
Error response from daemon: Get http://10.8.0.1:80/v2/: Get http://10.8.0.1./service/token?account=admin&client_id=docker&offline_token=true&service=harbor-registry: dial tcp: lookup 10.8.0.1.: no such host (Pay attention on the last dot after ip)

See in the docker-compose       
- HARBORHOSTNAME=${harborhostname}.${traefikdomain}

${traefikdomain} will be not set.

paulobezerr:
This variable creates an error in configuration files:

```
configadmin...env:2:EXT_ENDPOINT=http://domain.
data...config.json:14:  "ext_endpoint": "http://domain.",
registry...config.yml:24:    realm: http://domain./service/token
```

This makes UI works well, but you will never login with
docker login or pull something public.

Once this variable was never defined, I removed it.
Than works everything fine.

Closes #707

### DIFF
--- a/templates/project-harbor/4/docker-compose.yml
+++ b/templates/project-harbor/4/docker-compose.yml
@@ -47,7 +47,7 @@ services:
     image: mreferre/harbor-setupwrapper:1.1.1-1
     container_name: harbor-setupwrapper
     environment:
-      - HARBORHOSTNAME=${harborhostname}.${traefikdomain}
+      - HARBORHOSTNAME=${harborhostname}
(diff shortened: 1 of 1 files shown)


rawmind0:
Hi @paulobezerr , thanks for this PR.

LGTM

PR #772, opened 2018-04-12
es-cluster: fix imporper usage of sidekick volume

ExiaSR:
The current template will not work on `rancher/server` v1.6.14 above.
Also, a sidekick is not supposed to be used by multiple primary services.


rawmind0:
Hi @ExiaSR , thanks so much for this PR.

Sorry, i didn't see this PR before, but issue already solved by https://github.com/rancher/community-catalog/pull/789 and https://github.com/rancher/community-catalog/pull/790

ExiaSR:
@rawmind0 Sure, no worry :)

PR #788, opened 2018-05-17
fix/single-sidekick-issue

mattpallissard:
More than one service cannot reference the same sidekick.   Make multiple side kicks.
There is probably a clever way to do this, but it works.


rawmind0:
Superseded by https://github.com/rancher/community-catalog/pull/789

PR #789, opened 2018-05-18
Updated es-cluster package to work on rancher v1.6.15 and higher

rawmind0:
Due to merged PR https://github.com/rancher/rancher-compose-executor/pull/143, es-cluster is failing to deploy on rancher v1.6.15 and higher, issue https://github.com/rancher/community-catalog/issues/756

This PR rewrite 2 last es-cluster versions to be deployable on rancher v1.6.15 and higher and set `maximum_rancher_version: v1.6.14` on previous versions.

PR #790, opened 2018-05-21
Fix es-cluster package, split sysctl sidekicks for rancher v1.6.15 and higher

rawmind0:


rawmind0:
Fix sysctl sidekick for es-cluster package...Supplement https://github.com/rancher/community-catalog/pull/789

LGTM

PR #792, opened 2018-05-24
fix wordpress package lb

MariuszHenn:
According https://github.com/rancher/rancher/issues/8538 there is a bug in LB causes to generate invalid configuration when interval option is not set. This invalid config causes LB fail to reload/start.

It happens when upgrading wordpress load balancer to docker:rancher/lb-service-haproxy:v0.9.1.


rawmind0:
Hi @MariuszHenn .. Thanks for the PR...

LGTM...

PR #794, opened 2018-05-31
Update traefik package to v1.6.2

rawmind0:


rawmind0:
Tested and working....

LGTM

PR #801, opened 2018-06-12
Kafka package - Fix external ip advertise. Updated kafka to v1.1.0

rawmind0:


rawmind0:
Tested and working...

LGTM

Issue #803, opened 2018-06-16
Add easyappointments to cattle catalog


codejamninja:
I use easyappointments a lot and would like to add it as a cattle catalog item

rawmind0:
PR build is failing, please take a look at changes requested to address it.

PR #804, opened 2018-06-16
Added easyappointments catalog item to cattle

clayrisser:
Fixes issue #803

### DIFF
--- a/templates/easyappointments/0/docker-compose.yml
+++ b/templates/easyappointments/0/docker-compose.yml
@@ -1,0 +1,39 @@
+version: '2'
+
+services:
+  easyappointments:
+    image: codejamninja/easyappointments:${version}
+{{- if (.Values.public_port)}}
+    ports:
+      - ${public_port}:8888
+{{- end}}
+{{- if not (.Values.mysql_host)}}
+    links:
+      - mysql:db
+{{- end }}
+    environment:
+      BASE_URL: ${base_url}
+      DB_NAME: ${mysql_database}
+      DB_PASSWORD: ${mysql_password}
+      DB_USERNAME: ${mysql_user}
+{{- if (.Values.mysql_host)}}
+      DB_HOST: ${mysql_host}
+{{- end}}
+      TZ: ${timezone}
+    labels:
+      io.rancher.container.hostname_override: container_name
+
+{{- if not (.Values.mysql_host)}}
+  mysql:
+    image: mariadb:10.3.7
+    environment:
+      MYSQL_DATABASE: ${mysql_database}
+      MYSQL_USER: ${mysql_user}
+      MYSQL_PASSWORD: ${mysql_password}
+      MYSQL_RANDOM_ROOT_PASSWORD: true
+    volume_driver: ${volume_driver}
+    volumes:
+      - {{.Stack.Name}}--mysql-data:/var/lib/mysql
+    labels:
+      io.rancher.container.hostname_override: container_name
+{{- end}}
--- a/templates/easyappointments/0/rancher-compose.yml
+++ b/templates/easyappointments/0/rancher-compose.yml
@@ -1,0 +1,75 @@
+version: '2'
+
+catalog:
+  name: 'Easy!Appointments'
+  version: 0.1.0
+  description: 'Open source appointment scheduler'
+  questions:
+    - variable: base_url
+      label: 'Base URL'
+      type: string
+      required: true
+    - variable: timezone
+      label: Timezone
+      type: string
+      default: UTC
+      required: true
+    - variable: version
+      label: Version
+      type: enum
+      default: 1.3.0
+      options:
+        - 1.3.0
+        - 1.3
+        - 1
+        - 1.2.0
+        - 1.2
+        - latest
+    - variable: mysql_host
+      label: 'Mysql Host'
+      description: 'Leave empty to spin up a mysql docker container'
+      type: string
+      required: false
+    - variable: mysql_user
+      label: 'Mysql User'
+      type: string
+      default: easyappointments
+      required: true
+    - variable: mysql_password
+      label: 'Mysql Password'
+      type: password
+      required: true
+    - variable: mysql_database
+      label: 'Mysql Database'
+      type: string
+      default: easyappointments
+      required: true
+    - variable: volume_driver
+      label: 'Volume Driver'
+      required: true
+      default: rancher-nfs
+      type: enum
+      options:
+        - rancher-nfs
+        - rancher-efs
+        - rancher-ebs
+        - local
+    - variable: public_port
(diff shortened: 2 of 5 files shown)


templates/easyappointments/0/rancher-compose.yml
```diff
@@ -0,0 +1,75 @@
...
+        - 1.3
+        - 1
+        - 1.2.0
+        - 1.2
+        - latest
+    - variable: mysql_host
+      label: 'Mysql Host'
+      description: 'Leave empty to spin up a mysql docker container'
+      type: string
+      required: false
```
rawmind0: What do you think to replace `mysql_host` type string to type service??
```
    - variable: mysql_host
      label: 'Mysql Host'
      description: 'Leave empty to spin up a mysql docker container'
-      type: string
+      type: service
      required: false
```

templates/easyappointments/0/rancher-compose.yml
```diff
@@ -0,0 +1,75 @@
...
+      label: Version
+      type: enum
+      default: 1.3.0
+      options:
+        - 1.3.0
+        - 1.3
+        - 1
+        - 1.2.0
+        - 1.2
+        - latest
```
rawmind0: Use generic docker image tags, implicit or explicit, is a bad practice due to is not reproducible. Otherwise, catalog packages manage versions, update,... 

Could you please add one catalog package version for each docker image tag, one for `1.2.0` other for `1.3.0` and remove `latest`??

templates/easyappointments/0/rancher-compose.yml
```diff
@@ -0,0 +1,75 @@
+version: '2'
+
+catalog:
+  name: 'Easy!Appointments'
+  version: 0.1.0
```
rawmind0: What do you think if you match this version with service version??

```
-  version: 0.1.0
+  version: 1.3.0
```

templates/easyappointments/config.yml
```diff
@@ -0,0 +1,5 @@
+name: 'Easy!Appointments'
+description: |
+  Open source appointment scheduler
+version: 0.1.0
```
rawmind0: What do you think if you match this version with service version??

```
-version: 0.1.0
+version: 1.3.0
```

templates/easyappointments/0/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+      MYSQL_DATABASE: ${mysql_database}
+      MYSQL_USER: ${mysql_user}
+      MYSQL_PASSWORD: ${mysql_password}
+      MYSQL_RANDOM_ROOT_PASSWORD: true
+    volume_driver: ${volume_driver}
+    volumes:
+      - {{.Stack.Name}}--mysql-data:/var/lib/mysql
+    labels:
+      io.rancher.container.hostname_override: container_name
+{{- end}}
```
rawmind0: To use golang templating, you need to rename `templates/easyappointments/0/docker-compose.yml` to `templates/easyappointments/0/docker-compose.yml.tpl`

templates/easyappointments/0/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+      - mysql:db
+{{- end }}
+    environment:
+      BASE_URL: ${base_url}
+      DB_NAME: ${mysql_database}
+      DB_PASSWORD: ${mysql_password}
+      DB_USERNAME: ${mysql_user}
+{{- if (.Values.mysql_host)}}
+      DB_HOST: ${mysql_host}
+{{- end}}
```
rawmind0: What is the default value of env var `DB_HOST`?? What do you think to define like that??

```
{{- if not (.Values.mysql_host)}}
    links:
      - mysql:db
+{{- else }}
+    external_links:
+      - ${mysql_host}:db
{{- end }}
    environment:
      BASE_URL: ${base_url}
      DB_NAME: ${mysql_database}
      DB_PASSWORD: ${mysql_password}
      DB_USERNAME: ${mysql_user}
+      DB_HOST: db
-{{- if (.Values.mysql_host)}}
-      DB_HOST: ${mysql_host}
-{{- end}}
```

rawmind0 (changes requested):

rawmind0:
Hi @codejamninja . Thanks for the PR and contribute.

Could you please take a look to changes requested??

templates/easyappointments/0/rancher-compose.yml
codejamninja: I'm not familiar with type service

templates/easyappointments/0/rancher-compose.yml
codejamninja: Ok, sure

templates/easyappointments/0/rancher-compose.yml
codejamninja: Is that standard practice?

templates/easyappointments/0/docker-compose.yml
codejamninja: Ok, will do

templates/easyappointments/0/docker-compose.yml
codejamninja: That is so they can connect with a remote database if they so choose to. Read the description of the `msyql_host` question in the rancher config. Also, the default value of DB_HOST is db, hence the reason I link to db when mysql_host is not set.

templates/easyappointments/0/rancher-compose.yml
rawmind0: Trying to make it standard due to it's more clear.. Package version `1.3.0` install service version `1.3.0`

templates/easyappointments/0/rancher-compose.yml
rawmind0: type `service` shows combo box wit a list of stack/services running at rancher.. could be exposed at service as link or external_link... Useful when you have more than one ip to expose and it's resilient to service update/upgrade...

templates/easyappointments/0/docker-compose.yml
rawmind0: Yeah, i know what you are doing here and it works. Just a proposal that may be more clear.

templates/easyappointments/0/rancher-compose.yml
codejamninja: Um, yeah that's not how MySQL host is for. You only fill it out if you're doing an external connection

templates/easyappointments/0/docker-compose.yml
codejamninja: Ah, very nice

cjellick:
Closing this due to staleness. Feel free to reopen or open a new PR if there's still a relevant change to be made. Thanks!

PR #815, opened 2018-08-16
Updated to use v1.0.3

pdube:
Hi,

I have updated the version of the docker-machine-driver to use. 

Am I missing anything?

Thanks


pdube:
@jgreat

PR #834, opened 2019-03-05
Update sentry to 9.0.0

umarniz:
Update sentry from to 9.0.0.

Tested the upgrade in our infrastructure to work fine.

Tested the compose files (with dummy env variables) to create a new stack in rancher 1.6 and it worked fine.

### DIFF
--- a/templates/sentry/6/README.md
+++ b/templates/sentry/6/README.md
@@ -1,0 +1,30 @@
+# Sentry
+
+
+### Info:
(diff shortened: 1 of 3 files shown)


cjellick:
Sorry this hasn't gotten a response. Is it still relevant? Even though it's old, seems like it could be.
I'm going to close due to age, but feel free to reopen if you think it's still relevant. Thanks!

PR #840, opened 2019-09-08
Update to traefik 1.7.14

blar:
Update to traefik 1.7.14

PR #841, opened 2019-09-10
python tox installation is breaking the tests

blar:
Use version 3.13.2

https://github.com/rancher/community-catalog/pull/840#issuecomment-529867371


rawmind0:
My comment https://github.com/rancher/community-catalog/pull/840#issuecomment-529867371 was not fully right. The issue is related with python tox installation but not about the version. It seems that it needs to be installed in a separate process once pip is installed, in order to work properly.

PR #842, opened 2019-09-10
Update to traefik 1.7.14 #2

blar:
Update to traefik 1.7.14


blar:
@rawmind0 Now it works.

PR #847, opened 2020-04-12
Added terraria server

barchero:
Hello, i've added a Terraria server template in my custom rancher catalog and i think someone might be interested.

The template is using [beardedio/terraria](https://hub.docker.com/r/beardedio/terraria) docker images.


barchero:
i'm sorry.
only want to add the last commit.

i'm going to create another PR properly

